PR #1, opened 2017-09-19
Update scripts

gpetruc:
* move the postprocessor driver to `scripts/nano_postproc.py` and make it executable
 * add nano_report.py


arizzi:
should we also symlink haddnano.py to nano_hadd.py ?

PR #4, opened 2017-09-19
example of factorized jec uncertainty producer

arizzi:


arizzi:
it is quite slow as far as I can tell (~170 Hz)  likely the double loop can be written better, perhaps adding c++ helpers, plus everything is string based...

arizzi:
@gpetruc running without a cut string seem to crash the nano_postproc.py

python/postprocessing/examples/jecUncertProducer.py
```diff
@@ -0,0 +1,106 @@
...
+        pass
+    def analyze(self, event):
+        """process event, return True (go to next module) or False (fail, go to next event)"""
+        jets = Collection(event, "Jet")
+	for u in self.uncerts :
+	    uworker=self.factorizedUncertainties[u]
+	    jetUn=[]
+       	    for j in jets :
+        	uworker.setJetEta(j.eta)
+        	uworker.setJetPt(j.eta)
```
gpetruc: j.pt?

arizzi:
yeah sounds like a good idea !

gpetruc:
https://github.com/gpetruc/nanoAOD-tools/tree/bugfix_pp should fix it

Il 19 Set 2017 20:38, "arizzi" <notifications@github.com> ha scritto:

PR #7, opened 2017-09-20
Standalone environment setup

peruzzim:


standalone/env_standalone.sh
```diff
@@ -0,0 +1,26 @@
...
+	    echo "Build directory is already present, please source again standalone/env_standalone.sh without the build argument."
+	else
+	    find build/lib/python python -type d -execdir touch '{}/__init__.py' \;
+	    export NANOAODTOOLS_BASE=${PWD}
+	    export PYTHONPATH=${NANOAODTOOLS_BASE}/build/lib/python:${PYTHONPATH}
+	    echo "Standalone environment set."
+	fi
+    fi
+else
+    echo "Please cd to the NanoAODTools main directory before trying to source standalone/env_standalone.sh"
```
gpetruc: there should be a way to allow understanding in which directory the script is, to avoid this restriction (like 'thisroot.sh' does).
if it's easy can you do it?
otherwise we merge as is, and improve later

PR #9, opened 2017-09-20
Collection merger module

peruzzim:


python/postprocessing/examples/collectionMerger.py
```diff
@@ -0,0 +1,77 @@
...
+            self.out.branch("%s_%s"%(self.output,br), _rootLeafType2rootBranchType[self.branchType[br]], lenVar="n%s"%self.output)
+
+    def endFile(self, inputFile, outputFile, inputTree, wrappedOutputTree):
+        pass
+
+    def filterBranchNames(self,branches,collection):
+        out = []
+        for br in branches:
+            name = br.GetName()
+            if not re.match(collection+'_.*',name): continue
```
gpetruc: any reason why it's not `name.startswith(collection+"_")` ?

python/postprocessing/examples/collectionMerger.py
```diff
@@ -0,0 +1,77 @@
...
+            out.append(name.replace(collection+'_',''))
+            self.branchType[out[-1]] = br.FindLeaf(br.GetName()).GetTypeName()
+        return out
+
+    def analyze(self, event):
+        """process event, return True (go to next module) or False (fail, go to next event)"""
+        coll = [Collection(event,x) for x in self.input]
+        objects = [(coll[j][i],j,i) for j in xrange(self.nInputs) for i in xrange(len(coll[j]))]
+        if self.selector: objects=filter(lambda (obj,j,i) : self.selector(obj), objects)
```
gpetruc: I think it would make more sense to have that the selectors can be a `dict` with a key per collection, and the lambda be a function that takes the object (and the index, if you think it's useful... I can't see why)

python/postprocessing/examples/collectionMerger.py
```diff
@@ -0,0 +1,77 @@
...
+        branches = [_brlist.At(i) for i in xrange(_brlist.GetEntries())]
+        self.brlist_sep = [self.filterBranchNames(branches,x) for x in self.input]
+        self.brlist_all = set(itertools.chain(*(self.brlist_sep)))
+
+        self.is_there = np.zeros(shape=(len(self.brlist_all),self.nInputs),dtype=bool)
+        for bridx,br in enumerate(self.brlist_all):
+            for j in xrange(self.nInputs):
+                if br in self.brlist_sep[j]: self.is_there[bridx][j]=True
+
+        self.nullValue = [ 0.0 if self.branchType[br] in ['Float_t','Double_t'] else 0 for br in self.brlist_all]
```
gpetruc: parethesis around `(0.0 if ... else 0)` would make it more readable.
that said, it makes no difference between 0 and 0.0 since python & root will cast it to the right value anyway at the end, so it could just be all zeros and you can do without it

python/postprocessing/examples/collectionMerger.py
```diff
@@ -0,0 +1,77 @@
...
+        objects = [(coll[j][i],j,i) for j in xrange(self.nInputs) for i in xrange(len(coll[j]))]
+        if self.selector: objects=filter(lambda (obj,j,i) : self.selector(obj), objects)
+        objects.sort(key = self.sortkey, reverse = self.reverse)
+        if self.maxObjects: objects = objects[:self.maxObjects]
+        for bridx,br in enumerate(self.brlist_all):
+            out = []
+            nullvalue = self.nullValue[bridx]
+            for obj,j,i in objects:
+                val = getattr(obj,br) if self.is_there[bridx][j] else nullvalue
+                if type(val)==str: val = ord(val)
```
gpetruc: when do we get this? uint8s? 
we should probably fix the reader as we don't want them to become strings in the first place...

python/postprocessing/examples/collectionMerger.py
```diff
@@ -0,0 +1,77 @@
...
+                out.append(val)
```
gpetruc: could be worth checking if using a pre-allocated array of ints or floats will be better than doing append in a python list

PR #10, opened 2017-09-20
Add crab scripts, options to save provenance in postprocessing and create FJR 

arizzi:
Run post processing on crab.
Support lumi-splitting and eventsAwareLumiSplitting
runsAndLumis json can now be passed as a dict rather than json file

Also introduce options to create FrameworkJobReport out of the python postprocessor (used by the crab config)

### DIFF
--- a/python/postprocessing/framework/jobreport.py
+++ b/python/postprocessing/framework/jobreport.py
@@ -1,0 +1,57 @@
+import xml.etree.cElementTree as ET
+#import lxml.etree.ElementTree as ET
+class JobReport :
+       def __init__(self):
+	   self.fjr = ET.Element("FrameworkJobReport")
+           self.readbranches = ET.SubElement(self.fjr, "ReadBranches")
+           self.performancereport = ET.SubElement(self.fjr, "PerformanceReport")
+           self.performancesummary = ET.SubElement(self.performancereport, "PerformanceSummary", Metric="StorageStatistics")
+           ET.SubElement(self.performancesummary, "Metric", Name="Parameter-untracked-bool-enabled",Value="true")
+    #<Metric Name="Parameter-untracked-bool-enabled" Value="true"/>
+    #<Metric Name="Parameter-untracked-bool-stats" Value="true"/>
+    #<Metric Name="Parameter-untracked-string-cacheHint" Value="application-only"/>
+    #<Metric Name="Parameter-untracked-string-readHint" Value="auto-detect"/>
+    #<Metric Name="ROOT-tfile-read-totalMegabytes" Value="0"/>
+    #<Metric Name="ROOT-tfile-write-totalMegabytes" Value="0"/>
+
+#likely not needed
+#<GeneratorInfo>
+#</GeneratorInfo>
+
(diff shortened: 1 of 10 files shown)


python/postprocessing/framework/jobreport.py
```diff
@@ -0,0 +1,57 @@
...
+	   ET.SubElement(infile,"InputType").text="primaryFiles"
+	   ET.SubElement(infile,"ModuleLabel").text="source"
+	   ET.SubElement(infile,"InputSourceClass").text="PoolSource"
+	   ET.SubElement(infile,"GUID").text=""
+	   ET.SubElement(infile,"EventsRead").text="%s"%eventsRead
+	   runs=ET.SubElement(infile,"Runs")
+	   for r,ls in runsAndLumis.iteritems() :
+	   	run=ET.SubElement(runs,"Run",ID="%s"%r)
+	   	for l in ls :
+		   ET.SubElement(run,"LumiSection",ID="%s"%l)
```
gpetruc: these we could fill with the correct info easily

python/postprocessing/framework/jobreport.py
```diff
@@ -0,0 +1,57 @@
...
+
+       def addOutputFile(self,filename,events=1,runsAndLumis={"1":[1]}):
+           infile = ET.SubElement(self.fjr, "File")
+           ET.SubElement(infile,"LFN").text=""
```
gpetruc: `re.sub(r".*?/(/store/.*\.root)(\?.*)?","$1", filename)` ? 
(o via edmFileUtil)

python/postprocessing/framework/postprocessor.py
```diff
@@ -20,6 +22,12 @@ def __init__(self,outputDir,inputFiles,cut=None,branchsel=None,modules=[],compre
 	self.noOut=noOut
 	self.friend=friend
 	self.justcount=justcount
+	self.provenance=provenance
+	self.jobReport = JobReport() if fwkJobReport else None
+	self.haddFileName=haddFileName
+	if self.jobReport and not self.haddFileName :
+		print "Because you requested a FJR we assume you want the final hadd. No name specified for the output file, will use tree.root"
+		self.haddFileName="tree.root"
```
gpetruc: io forse qui metterei throw

python/postprocessing/framework/jobreport.py
arizzi: yes, also the total number of events and other info can be corrected

python/postprocessing/framework/jobreport.py
arizzi: it doesn't matter this one, it is anyhow moved later by crab

python/postprocessing/framework/postprocessor.py
arizzi: actually tree.root is the name the crab config expects, so it is fine to have a (this) default

arizzi:
should be working now

it can be merged

gpetruc:
Will you also update the FJR in order to put meaningful info on the number of events and run/lumi content of the files, or that will come later?

arizzi:
definitelly later...

PR #11, opened 2017-09-21
Pileup weights

emanueledimarco:
- weights for the time being computed with number of vertices (should be updated with nTrue interactions)
- protection for large weights ON by default
- uses two input distributions for MC (Summer16 profile and full 2016 data), but are just examples


emanueledimarco:
- updated to use the number of true interactions (needs [this PR in the nanoAOD](https://github.com/cms-nanoAOD/cmssw/pull/12) )
- it checks if the variable exists in the tree, otherwise puts 1 (e.g. when running on data)

arizzi:
can you check how fast does it run?

emanueledimarco:
"By eye" it is >10 kHz on the nano.root produced on 10k events the usual ttbar MC file.
Is there a std way to check the speed, not considering the begin / end job time?

PR #15, opened 2017-09-22
Expose values calculated by previous eventLoop modules

peruzzim:
Defines a dictionary in InputTree keeping newly added branches.
This is checked before looking for (or creating) any TTreeReader in treeReaderArrayTools.readBranch.
The dictionary is cleaned when starting to process each event in eventLoop.


arizzi:
@peruzzim can you have a look at the conflict?

PR #16, opened 2017-09-25
C++ helper module support for post-proc looper

peruzzim:


python/postprocessing/framework/treeReaderArrayTools.py
```diff
@@ -21,31 +21,30 @@ def InputTree(tree,entrylist=None):
...
        if not leaf.GetLen() == 1: raise RuntimeError, "Branch %s is not a value" % branchName
        typ = leaf.GetTypeName()
-       tree._ttrvs[branchName] = _makeValueReader(tree, typ, branchName, remakeAllFirst=not(isClean))
+       tree._ttrvs[branchName] = _makeValueReader(tree, typ, branchName)
     return tree._ttrvs[branchName]
 
 
 def readBranch(tree, branchName):
     """Return the branch value if the branch is a value, and a TreeReaderArray if the branch is an array"""
+    if tree._ttreereader._isClean: raise RuntimeError, "readBranch must not be called before setting the TTreeReader entry (or calling Next())"
```
gpetruc: I think the error message here should tell the user to call `gotoEntry` (you don't want him to call `SetEntry` directly, behind your back; and, it even if he did, he'd still get this exception since you're checking `_isClean`)

src/mhtjuProducerCppWorker.cc
```diff
@@ -0,0 +1,15 @@
...
+
+    math::XYZTLorentzVectorF ht(0,0,0,0);
+
+    unsigned n = (*nJet).Get()[0];
+    for (unsigned i=0; i<n; i++){
+      ht += math::PtEtaPhiMLorentzVectorF((*Jet_pt)[i],0,(*Jet_phi)[i],0);
+    }
+
+    std::vector<float> result = {ht.Pt(),ht.Phi()};
+    return result;
```
gpetruc: better return a fixed-size object, e.g. `std::pair<float,float>` or even just the `XYZTLorentzVectorF`

python/postprocessing/examples/mhtjuProducerCpp.py
```diff
@@ -0,0 +1,52 @@
...
+from PhysicsTools.NanoAODTools.postprocessing.framework.eventloop import Module
+
+class mhtjuProducerCpp(Module): # MHT producer, unclean jets only (no lepton overlap cleaning, no jet selection)
+    def __init__(self):
+        if "/mhtjuProducerCppWorker_cc.so" not in ROOT.gSystem.GetLibraries():
+            print "Load C++ mhtjuProducerCppWorker worker module"
+            base = os.getenv("NANOAODTOOLS_BASE")
+            if not base:
+                base = "%s/src/PhysicsTools/NanoAODTools"%os.getenv("CMSSW_BASE")
+            ROOT.gROOT.ProcessLine(".L %s/src/mhtjuProducerCppWorker.cc+O"%base)
```
gpetruc: if the code is compiled beforehand, one should load the library instead of loading the .cc

interface/mhtjuProducerCppWorker.h
```diff
@@ -0,0 +1,28 @@
...
+#include <vector>
+#include <TTreeReaderValue.h>
+#include <TTreeReaderArray.h>
+#include "DataFormats/Math/interface/LorentzVector.h"
+
+class mhtjuProducerCppWorker {
+
+public:
+
+  mhtjuProducerCppWorker(){};
```
gpetruc: no ";"
also, better initialize the pointers (to nullptr)
(this is an example, so it should be better written very cleanly)

python/postprocessing/examples/mhtjuProducerCpp.py
```diff
@@ -0,0 +1,52 @@
+import ROOT
+import os
+ROOT.PyConfig.IgnoreCommandLineOptions = True
+
+from PhysicsTools.NanoAODTools.postprocessing.framework.datamodel import Collection
```
gpetruc: not used

arizzi:
should we merge this?

arizzi:
guys, are we happy to merge this or not?

PR #22, opened 2017-10-24
added data/MC scale factors (per jet weights) for b-tag efficiency an…

veelken:
…d mistag rate


arizzi:
@veelken did you check the speed on a largish sample (so that we avoid being dominated by the startup time) ?
How fast does it go?

veelken:
@arizzi The code is running very fast (I ran 1000 events), but I don't have timing numbers. Is there an option to get timing information ?

arizzi:
they should be printed out automatically by the the postprocessing fwk...
just try more events (more than 10k)

On Tue, Oct 24, 2017 at 9:24 PM, Christian Veelken <notifications@github.com

veelken:
Hi Andrea,

the eventloop.py file has a small logic error: in case one runs over 10k events (all I have), you don't get the timing output, because the timing output is printed when the 10001st event gets processed. After changing progress to 5000 events, I get the timing output attached below [*].
This is for a ttbar MC sample.

Is 500 processed events per second ok ?

Cheers,


Christian

- - - - - - - -

[*]

Processed     5000/   10000 entries (elapsed time    10.5s, curr speed    0.474 kHz, avg speed    0.474 kHz), accepted     5001/    5001 events (100.00%)
Processed 10000 entries from /hdfs/local/karl/nanoAOD/92X_upgrade2017_realistic_v10_ext1-v1/TT_1.root, selected 10000 entries

peruzzim:
@veelken the CMG version runs at 1.7 kHz with one "working point" (the full shape reweighting), so it sounds reasonable to me.

Maybe it could be made faster using a C++ helper (requires cms-nanoAOD/nanoAOD-tools#16), but from previous tests of mine it looked like most of the time was spent in actually evaluating the scale factor because of internal interpolations, rather than on data access.

Could you please try to make a test, setting sf = 1 instead of calling eval_auto_bounds in these lines:
https://github.com/veelken/nanoAOD-tools/blob/4ae2bfee7495998992839dc6b973caffb5037561/python/postprocessing/examples/btagSFProducer.py#L181-L185
so we can factorize the two contributions?

And please, let's also look at the next 5k events to be sure there is no init time erroneously accounted in the benchmark report.

veelken:
Hi,

with using BTagCalibrationReader::eval_auto_bounds I get
   Processed     4999/   10000 entries (elapsed time    10.5s, curr speed    0.476 kHz, avg speed    0.476 kHz), accepted     5000/    5000 events (100.00%)
Processed     9998/   10000 entries (elapsed time    29.4s, curr speed    0.265 kHz, avg speed    0.341 kHz), accepted     9999/    9999 events (100.00%)
and when always returning 1 for the SF (and not using the BTagCalibrationReader::eval_auto_bounds function) I get
   Processed     4999/   10000 entries (elapsed time     9.6s, curr speed    0.520 kHz, avg speed    0.520 kHz), accepted     5000/    5000 events (100.00%)
Processed     9998/   10000 entries (elapsed time    27.1s, curr speed    0.286 kHz, avg speed    0.369 kHz), accepted     9999/    9999 events (100.00%)

It seems to me that most of the time is spend on reading the input data.

To confirm this hypothesis, I just load the input data (i.e. only executing the statement "jets = Collection(event, 'Jet')") in the analyze function of the btagSFProducer class), and I get
   Processed     4999/   10000 entries (elapsed time     5.1s, curr speed    0.986 kHz, avg speed    0.986 kHz), accepted     5000/    5000 events (100.00%)
Processed     9998/   10000 entries (elapsed time    16.8s, curr speed    0.427 kHz, avg speed    0.596 kHz), accepted     9999/    9999 events (100.00%),
i.e. most of the time is indeed spend on loading the input data. The remaining time is spend on iterating over the jet collection x systematic uncertainties.

Ok, I think I did my fair share of timing studies :)

Can the PR please be merged now ?

arizzi:
was the file local or remote? the reading speed should be much better than this
Were you running only the SF or also the JECs?

Anyhow I'll do some checks myself too.

PR #23, opened 2017-10-27
fixed bugs that caused btagSF to be evaluated to zero for some jets:

veelken:
- added workaround for the fact that BTagCalibrationReader::eval_auto_bounds() returns zero
  in case jet pT > 30 or abs(eta) > 2.4
- fixed processing of btagSF for charm and bottom jets

### DIFF
--- a/python/postprocessing/modules/btv/btagSFProducer.py
+++ b/python/postprocessing/modules/btv/btagSFProducer.py
@@ -31,7 +31,7 @@ class btagSFProducer(Module):
     """Calculate btagging scale factors
         algo has to be either 'csvv2' or 'cmva'
     """
-    def __init__(self, algo = 'csvv2', verbose = 0):
+    def __init__(self, algo = 'csvv2', verbose = 1):
 
         self.algo = algo.lower()
 
@@ -156,8 +156,8 @@ class btagSFProducer(Module):
             If unknown wp/syst/mtype/flavor, returns -1.0
         """
 
-        if not (pt > self.min_pt and pt < self.max_pt and abs(eta) < self.max_abs_eta):
-            return 1.
+        ##if not (pt > self.min_pt and pt < self.max_pt and abs(eta) < self.max_abs_eta):
+        ##    return 1.
 
         flavor_btv = None
         if abs(flavor) == 5:
--- a/python/postprocessing/modules/btv/btagSFProducer.py
+++ b/python/postprocessing/modules/btv/btagSFProducer.py
@@ -38,13 +38,8 @@ class btagSFProducer(Module):
         self.verbose = verbose
 
         # CV: Return value of BTagCalibrationReader::eval_auto_bounds() is zero
-        #     in case no SF is defined for a given jet pT and eta in the btagSF.csv file !!
-        #
-        #     In the current version of the btagSF.csv files this is the case for jet pT < 30 GeV or abs(eta) > 2.4
-        #
+        #     in case jet abs(eta) > 2.4 !!
         self.max_abs_eta = 2.4
-        self.min_pt = 30.
-        self.max_pt = 670.
 
         # define measurement type for each flavor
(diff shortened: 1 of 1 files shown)


arizzi:
this is not coherent with what stated here
https://twiki.cern.ch/twiki/bin/view/CMS/BTagCalibration
namely:
In June 2016, starting at CMSSW_8_0_12 and CMSSW_8_1_0, the eval_auto_bounds function is added, providing access to all loaded sys. types and automatically increasing the uncertainty by a factor of two, if the pt value is outside the provided range

arizzi:
are you sure it is both pt and eta out of range that gives zero? it should be only eta.

veelken:
Hi Andrea,

I have checked that the eval_auto_bounds function returns zero in case pT < 30 GeV or abs(eta) > 2.4. I will send an email to the BTV Hypernews and ask the author of that code to confirm.

veelken:
Here is the link to the email : https://hypernews.cern.ch/HyperNews/CMS/get/btag/1478.html

In most of the cases where the SF returns zero, the abs(eta) of the jet is indeed above 2.4. The case where the SF returns zero and the jet pT is below 30 GeV seems more rare (but does occur).

arizzi:
is the discriminator outside 0,1 boundary?

Il 30 ott 2017 9:30 AM, "Christian Veelken" <notifications@github.com> ha
scritto:

veelken:
Nope. It may be, however, that only in some bins of the discriminator distribution the SF evaluates to zero for jets of pT < 30 GeV.

python/postprocessing/modules/btv/btagSFProducer.py
```diff
@@ -184,37 +192,42 @@ def getSF(self, pt, eta, flavor, syst = 'central', wp = 'M', measurement_type =
...
+
+        for central_or_syst in self.central_and_systs:
             scale_factors = []
-            for central_or_syst in self.central_and_systs:
-                sf = self.getSF(jet.pt, jet.eta, jet.partonFlavour, central_or_syst, 'M', 'auto', False, discr)
-                #print " %s = %1.3f" % (self.branchNames[central_or_syst], sf)
+            for idx, jet in enumerate(jets):
+                sf = self.getSF(jet.pt, jet.eta, jet.partonFlavour, central_or_syst, 'M', 'auto', False, getattr(jet, discr))
+                if self.verbose > 0 and sf < 0.01:
+                    print("jet #%i: pT = %1.1f, eta = %1.1f, discr = %1.3f, flavor = %i" % (idx, jet.pt, jet.eta, getattr(jet, discr), jet.partonFlavour))
```
arizzi: here I would add if sf < 0.01 then sf =1

python/postprocessing/modules/btv/btagSFProducer.py
```diff
@@ -184,37 +192,42 @@ def getSF(self, pt, eta, flavor, syst = 'central', wp = 'M', measurement_type =
...
+            #print(self.branchNames_central_and_systs[central_or_syst], scale_factors)
+        for central_or_syst in self.central_and_systs_shape_corr:
             scale_factors = []
-            for central_or_syst in self.central_and_systs_shape_corr:
-                sf = self.getSF(jet.pt, jet.eta, jet.partonFlavour, central_or_syst, 'shape_corr', 'auto', True, discr)
-                #print " %s = %1.3f" % (self.branchNames[central_or_syst], sf)
+            for idx, jet in enumerate(jets):
+                sf = self.getSF(jet.pt, jet.eta, jet.partonFlavour, central_or_syst, 'shape_corr', 'auto', True, getattr(jet, discr))
+                if self.verbose > 0 and sf < 0.01:
+                    print("jet #%i: pT = %1.1f, eta = %1.1f, discr = %1.3f, flavor = %i" % (idx, jet.pt, jet.eta, getattr(jet, discr), jet.partonFlavour))
```
arizzi: also here

veelken:
Hi Andrea,

I have updated the code following your requests and what we discussed on Skype. Please have a look at the code again and merge PR #23 if everything is ok.

PR #26, opened 2017-11-16
HepMC from NanoAOD

jhgoh:
This PR adds an example worker module which can reproduce HepMC from the GenParticles in the NanoAOD. This module writes HepMC to a separate file in the standard HepMC ascii format.

```
python scripts/nano_postproc.py -s suffix --friend \
  -I PhysicsTools.NanoAODTools.postprocessing.modules.common.hepmcDump hepmcDump \
  OUTPUT_DIRECTORY INPUT_ROOT_FILE
```
produces hepmc.dat under your working directory.
The output can be re-analyzed using other softwares.

Caveats: this does not 100% reproduce the HepMC due to missing informations in the NanoAOD and round off.
 - Some information like Q^2, QCD/QED scales are set to default
 - Incident parton's pz, energy can be inaccurate because recomputing pz from GenPart_eta (typically > 10000) diverges. pz is re-estimated using x1,x2 and cm energy
 - Any GenParticles without mother are attached to the production vertex.


arizzi:
can you move the py module from examples to modules/common ?

jhgoh:
@arizzi done.

PR #32, opened 2017-12-06
Proposal for interface to add ROOT objects like TH1

rappoccio:
Here is the interface for users to add their own ROOT objects to a persistent file. The use case is expected to be TH1* or equivalent. The user would do:

```
self.addObject( ROOT.TH1F('h_vpt',   'h_vpt',   100, 0, 1000) )
```


arizzi:
@rappoccio it is not clear to me how this should work... I mean "beginJob" is reimplemented in derived classes, so why should we do the booking there rather than in dedicated methods that are not reimplemented?

Of course one can do something like
<pre>
    def beginJob(self,histFile=None,histDirName=None):
        Module.beginJob(self,histFile,histDirName)

        self.h_vpt=ROOT.TH1F('sumpt',   'sumpt',   100, 0, 1000)
        self.addObject(self.h_vpt )


</pre>

rappoccio:
@arizzi yes I was expecting people to call the "Module.beginJob". I guess I was thinking polymorphically similar to the EDAnalyzer framework.

arizzi:
I added an example

rappoccio:
Ahh yes, I missed that. Thanks.

PR #39, opened 2018-01-31
fix to get all-python JEC/JER implementation running

scooperstein:
Small change in implementation so that all-python JEC/JER implementation works with preliminary post-processor implementation.


arizzi:
@rappoccio are you still working on getting all the JEC/JER/Smearing/MET consistent right?
would it be ok if we temporarly solve the issue we had with smearing + JECU merging this one?

rappoccio:
Yes, still working on it, sorry. Go ahead with this one.

PR #40, opened 2018-02-05
Fix reading JES uncertainties file

ktht:
When trying to run jetmetUncertainties module https://github.com/cms-nanoAOD/nanoAOD-tools/blob/271cf2c755c02292e7c64cafc720eb313ed9ce3a/python/postprocessing/modules/jme/jetmetUncertainties.py#L272 the JetCorrectorParameters instance is constructed from the file name `$CMSSW_BASE/src/PhysicsTools/NanoAODTools/data/jme/Summer16_23Sep2016V4_MC_Uncertainty_AK4PFchs.txt` and `'Total'` as the uncertainty source/label https://github.com/cms-nanoAOD/nanoAOD-tools/blob/271cf2c755c02292e7c64cafc720eb313ed9ce3a/python/postprocessing/modules/jme/jetmetUncertainties.py#L57 This fails because the input file does not contain the label name between the brackets, which is expected because the uncertainty source label is provided for `JetCorrectorParameters`.


arizzi:
aren't those file automatically dump from with  cmsRun python config?

I'm not sure it makes sense to change the txt file ... can we make a solution where we fix the reader (or the dumper)

ktht:
Ok, makes sense. Preparing new PR in a moment.

Issue #45, opened 2018-02-15
update btagSF


veelken:
Update https://github.com/cms-nanoAOD/nanoAOD-tools/blob/master/python/postprocessing/modules/btv/btagSFProducer.py and related .csv files with corrections to BTV POG recipe on twiki https://twiki.cern.ch/twiki/bin/view/CMS/BtagRecommendation94X .

I volunteer to start working on this if you agree.

arizzi:
thanks. we should add new files, without replacing existing ones.
Ideally we should make things more configurable then they are (right now the tagger to filename association is hardcoded)

veelken:
Hi Andrea,

can you confirm that the branch Jet_btagCMVA in the nanoAOD Ntuple corresponds to the DeepCSV tagger ?

arizzi:
nope, DeepB is DeepCSV
https://github.com/cms-nanoAOD/cmssw/blob/master/PhysicsTools/NanoAOD/python/jets_cff.py#L130

veelken:
The new b-tagging scale factors for 2017 ReReco data and RunIIFall17 MC are implemented in the PR https://github.com/cms-nanoAOD/nanoAOD-tools/pull/46 . I also made the name of the CSV file with the scale factors configurable. Note that the shape corrections ("iterativefit" method) is not provided by the BTV POG for the 2017 ReReco data and RunIIFall17 MC yet, so the branches Jet_btagSF_shape_* are all 1.0 at the moment (only the branches Jet_btagSF, Jet_btagSF_up, and Jet_btagSF_down are filled with "proper" values so far).

Issue #33, opened 2017-12-18
Typo in btagSFProducer


ktht:
I believe there's a typo:
https://github.com/cms-nanoAOD/nanoAOD-tools/blob/7e661bf1e09f9d08541e445aedfdbd4cbbe83119/python/postprocessing/modules/btv/btagSFProducer.py#L200
Instead, it should be:
```python
            discr = "btagCSVV2"
```
(Sorry, didn't bother creating a PR for this.)

veelken:
The bug is fixed in PR https://github.com/cms-nanoAOD/nanoAOD-tools/pull/46 . This issue can be closed now.

PR #46, opened 2018-02-15
- updated btagSF to 2017 ReReco data and RunIIFall17 MC (adressing is…

veelken:
…sue https://github.com/cms-nanoAOD/nanoAOD-tools/issues/45)

- fixed bug reported in issue https://github.com/cms-nanoAOD/nanoAOD-tools/issues/33


python/postprocessing/modules/btv/btagSFProducer.py
```diff
@@ -43,19 +43,31 @@ def __init__(self, algo = 'csvv2', verbose = 0):
...
+            if self.inputFileName is None:
+                self.inputFileName = "DeepCSV_94XSF_V1_B_F.csv"
+            print("Loading btagSF weights for deep-CSV (b) algorithm from file '%s'" % os.path.join(self.inputFilePath, self.inputFileName))
+            self.measurement_types = {
+                0 : "comb",  # b
+                1 : "comb",  # c
+                2 : "incl"   # light
+            }
         elif self.algo == "cmva":
-            self.inputFileName = "btagSF_cMVAv2_ichep2016.csv"
```
arizzi: this is changing the default behaviour.

Can you please make it such 

-btagSF() still creates 2016 SF while (or remove the btagSF() and create a btagSF2016() )
-btagSF2017() has default config for 2017.

python/postprocessing/modules/btv/btagSFProducer.py
```diff
@@ -107,7 +119,9 @@ def beginJob(self):
         # (cf. https://twiki.cern.ch/twiki/bin/viewauth/CMS/BTagCalibration )
         self.calibration = ROOT.BTagCalibration(self.algo, os.path.join(self.inputFilePath, self.inputFileName))
         self.readers = {}
-        for wp in [ "L", "M", "T", "shape_corr" ]:
+        #for wp in [ "L", "M", "T", "shape_corr" ]:
+        # CV: shape corrections ('iterativefit' method) not supported for 2017 ReReco data and RunIIFall17 MC yet
+        for wp in [ "L", "M", "T" ]:
```
arizzi: this is also a problem for 2016 processing, please make it so that it works as it worked before when running on 2016

python/postprocessing/modules/btv/btagSFProducer.py
veelken: I added btagSF2016() and btagSF2017() and removed btagSF()

python/postprocessing/modules/btv/btagSFProducer.py
veelken: Ok, done.

veelken:
Hi Andrea,

I have updated the PR according to your comments.

arizzi:
mhh wouldn't be better to pass the configuration dict instead of just passing "2016" and have the dict hardcoded?
perhaps we could also build the dict in a less cut&paste way, i.e. with a default preferred config and override for methods that are not yet available...

veelken:
Hmh, I don't see that moving the dictionary to global scope would improve anything in regard to readibility of the code. But please go ahead and change it as you prefer (I think it will be easier to change the code directly than to discuss the change in detail). Ok ?

veelken:
Hi,

would be great if this PR can be merged soon.

Cheers,

Christian

PR #47, opened 2018-02-15
Move PU weight producer to module

jpata:
Moved PU weight producer to separate module, as ACLIC causes problems when running multiple jobs from the same working directory by overwriting/cleaning the temporary object files. Maybe there is a way to set the ACLIC compilation directory, but this seems cleaner.

### DIFF
--- a/interface/WeightCalculatorFromHistogram.h
+++ b/interface/WeightCalculatorFromHistogram.h
@@ -1,0 +1,34 @@
+#ifndef PhysicsTools_NanoAODTools_WeightCalculatorFromHistogram_h
+#define PhysicsTools_NanoAODTools_WeightCalculatorFromHistogram_h
+
+#include <iostream>
+#include <iomanip>
+#include <vector>
+#include <algorithm>
+#include <TH1.h>
+
+class WeightCalculatorFromHistogram {
+ public:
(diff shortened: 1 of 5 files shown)


jpata:
Probably it needs a fix in the loading path for crab.

arizzi:
I do not think we need to move it around... the problem you have is with the ++ that force to recompile which is on the other hand needed on crab.

So perhaps the clean way is to find a way that works both on crab and local. Probably just a try/except to first attempt loading without recompiling and then triggering the recompile on fail could work

arizzi:
looking better at your code... perhaps on crab it works because the libraries are passed. Can you please check? (the ACLIC line would not work instead because src is not passed to crab)

jpata:
I was also thinking now that it actually should directly work on crab due to the `try/except`. Or the idea you propose previously

I was thinking if I move the module, then it is normally compiled using CMSSW using `scram b`, such that errors can be found in integration builds etc. The code could stay basically the same, just a different location. Also I'm somewhat biased against ACLIC and dynamic compilation.

arizzi:
can you give it a try on crab?

jpata:
yes, I'm doing it.

jpata:
Yep, it works also in crab.

PR #48, opened 2018-02-16
Generalizing tools to enable AK8

rappoccio:
Here are a few tools that needed to be updated to be able to use AK8 jets with the same software as we have.

Currently this is not very performant. A C++ worker + python wrapper class is still necessary. Working on that part next, but this at least works correctly.


arizzi:
@veelken can you have a look at this PR (given that you also have a PR on this code)

veelken:
For me, the changes in this PR look fine (but Sal has reorganized the code quite a bit, which makes the comparison a bit difficult)

arizzi:
I propose I merge this and then you check you PR against it fixing when/if needed.. ok?

veelken:
Hi Sal,

the jetmetUncertainties tool included in this PR does not work. The problem is you have renamed some variables, but forgot to update all places where these variables are used. The error is attached below [*]. The problem is fixed in PR #50.

In the future, please do not forget to test your changes before submitting a PR, by running:
  python scripts/nano_postproc.py . /home/veelken/CMSSW_9_4_4/src/tthAnalysis/NanoAOD/test/nano.root -I PhysicsTools.NanoAODTools.postprocessing.modules.jme.jetmetUncertainties jetmetUncertainties2017 
(or similar)

Cheers,


Christian



[*]

  File "scripts/nano_postproc.py", line 46, in <module>
    p.run()
  File "/home/veelken/CMSSW_9_4_4/python/PhysicsTools/NanoAODTools/postprocessing/framework/postprocessor.py", line 122, in run
    (nall, npass, timeLoop) = eventLoop(self.modules, inFile, outFile, inTree, outTree)
  File "/home/veelken/CMSSW_9_4_4/python/PhysicsTools/NanoAODTools/postprocessing/framework/eventloop.py", line 60, in eventLoop
    ret = m.analyze(e) 
  File "/home/veelken/CMSSW_9_4_4/python/PhysicsTools/NanoAODTools/postprocessing/modules/jme/jetmetUncertainties.py", line 267, in analyze
    self.jesUncertainty[jesUncertainty].setJetPt(jet_pt_ref)
NameError: global name 'jet_pt_ref' is not defined

rappoccio:
Hmm, I had tested this with another module. I must have missed this one, sorry about that.

PR #51, opened 2018-02-22
improving speed of btag scale factor retrieval

kreczko:
# Summary
Dear all,

This PR improves the performance of the btag scale factor producer, `python/postprocessing/modules/btv/btagSFProducer.py`, by 40% (wall time).

# The issue this PR solves
A colleague of mine, @shane-breeze, observed that computing the btag scale factor does take a lot of time:
```
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    43354   31.869    0.001  140.315    0.003 btagSFProducer.py:249(analyze)
  7991874   63.137    0.000   76.069    0.000 btagSFProducer.py:196(getSF)
```

For our test sample this resulted in a total of
```
126005017 function calls (122184427 primitive calls) in 191.403 seconds
```

Looking at the code, we've found some improvement possibilities, the easiest ones are contained in this PR.

With the changes in place, the new profiling information is
```
53533045 function calls (49712455 primitive calls) in 116.827 seconds
    43354    9.206    0.000   66.927    0.002 btagSFProducer.py:284(analyze)
  8945662   34.541    0.000   37.679    0.000 btagSFProducer.py:229(getSFs)
```
The number correspond to a speedup of 40% and a reduction in function calls of 58% (calls are expensive after all.

# Implementation details
The implementation was done in several steps after each one the results were compared.
For simplicity, let's define `M = self.central_and_systs * njets + self.central_and_systs_shape_corr * njets` and `N_S = self.central_and_systs + self.central_and_systs_shape_corr`

## jet access
Previously jet variables were loaded `M` times per event. Now they are loaded once per event:
```
preloaded_jets = [(jet.pt, jet.eta, jet.partonFlavour, getattr(jet, discr)) for jet in jets]
```
This change alone bring gains of 16%.

## btag SF reader access
Previously, the btag SF reader was loaded M times per event. However, looking at the code this can be reduced to 2 times per event. Since the working points look static across events, it might be even worth moving the definitions into `beginJob`.

## other small things
 - moved `syst.lower()` outside `getSF` (`M` calls &rarr; `N_S` calls)
 - moved `flavor_btv` conversion outside `getSF` ( `M` calls &rarr; `njets` calls)

I've also moved the jet loop into the new `getSFs` function for readability.

### DIFF
--- a/python/postprocessing/modules/btv/btagSFProducer.py
+++ b/python/postprocessing/modules/btv/btagSFProducer.py
@@ -107,7 +107,7 @@ class btagSFProducer(Module):
                 raise ValueError("ERROR: Algorithm '%s' not supported for era = '%s'! Please choose among { %s }." % (self.algo, self.era, supported_algos))
         else:
             raise ValueError("ERROR: Algorithm '%s' not supported for era = '%s'! Please choose among { %s }." % (self.algo, self.era, supported_algos))
-        
+
         algoLabel = None
         if self.algo == "csvv2":
             algoLabel = "CSV (v2)"
@@ -162,7 +162,7 @@ class btagSFProducer(Module):
         # (cf. https://twiki.cern.ch/twiki/bin/viewauth/CMS/BTagCalibration )
         self.calibration = ROOT.BTagCalibration(self.algo, os.path.join(self.inputFilePath, self.inputFileName))
         self.readers = {}
-        for wp in self.supported_wp: 
+        for wp in self.supported_wp:
             wp_btv = { "l" : 0, "m" : 1, "t" : 2, "shape_corr" : 3 }.get(wp.lower(), None)
             syts = None
             if wp_btv in [ 0, 1, 2 ]:
@@ -193,25 +193,26 @@ class btagSFProducer(Module):
     def endFile(self, inputFile, outputFile, inputTree, wrappedOutputTree):
         pass
 
-    def getSF(self, pt, eta, flavor, syst = 'central', wp = 'M', measurement_type = 'auto', shape_corr = False, discr = 0.):
-        """Evaluate the SFs.
-            Note the flavor convention: hadronFlavor is b = 5, c = 4, f = 0
-            Convert them to the btagging group convention of 0 ,1, 2
-
-            Same for working points: input is 'L', 'M', 'T'
-            Convert to 0, 1, 2
-
-            Automatically checks if values are in allowed range
-
-            If unknown wp/syst/mtype/flavor, returns -1.0
+    def getReader(self, wp, shape_corr = False):
         """
+            Get btag scale factor reader.
+            Convert working points: input is 'L', 'M', 'T', 'shape_corr' to 0, 1, 2, 3
+        """
+        if shape_corr:
+            wp = "shape_corr"
+        wp_btv = { "l" : 0, "m" : 1, "t" : 2, "shape_corr" : 3 }.get(wp.lower(), None)
+        if wp_btv == None or not wp_btv in self.readers.keys():
+            if self.verbose > 0:
+                print("WARNING: Unknown working point '%s', setting b-tagging SF to -1!" % wp)
+            return -1.
+        return self.readers[wp_btv]
 
-        epsilon = 1.e-3
-        if eta <= -self.max_abs_eta:
-            eta = -self.max_abs_eta + epsilon
-        if eta >= +self.max_abs_eta:
-            eta = +self.max_abs_eta - epsilon
-
+    def getFlavorBTV(self, flavor):
+        '''
+            Maps hadronFlavor to BTV flavor:
+            Note the flavor convention: hadronFlavor is b = 5, c = 4, f = 0
+            Convert them to the btagging group convention of 0, 1, 2
+        '''
         flavor_btv = None
         if abs(flavor) == 5:
             flavor_btv = 0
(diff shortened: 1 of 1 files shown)


shane-breeze:
:+1:

python/postprocessing/modules/btv/btagSFProducer.py
```diff
@@ -193,25 +193,26 @@ def beginFile(self, inputFile, outputFile, inputTree, wrappedOutputTree):
...
+            Get btag scale factor reader.
+            Convert working points: input is 'L', 'M', 'T', 'shape_corr' to 0, 1, 2, 3
+        """
+        if shape_corr:
+            wp = "shape_corr"
+        wp_btv = { "l" : 0, "m" : 1, "t" : 2, "shape_corr" : 3 }.get(wp.lower(), None)
+        if wp_btv == None or not wp_btv in self.readers.keys():
+            if self.verbose > 0:
+                print("WARNING: Unknown working point '%s', setting b-tagging SF to -1!" % wp)
+            return -1.
```
ktht: Not a maintainer of this project, just wanted to point out that if there's no WP available for a particular b-tagging discriminator (e.g. shape corrections are missing for deep CSV at the time of writing), then this function returns a float but in successful cases it returns a reader object. It's going to raise an exception in `getSFs()` if the WP is not available, whereas the current (and IMO preferable) behavior is that we set the SFs to 1 in such cases. You can view and maybe cherry-pick my fix into your branch that is subject to this PR: https://github.com/HEP-KBFI/nanoAOD-tools/commit/174af42506dcbeed1558a9f4a34b4c40e1d2e998

kreczko: That is the same behaviour as is [currently implemented](https://github.com/cms-nanoAOD/nanoAOD-tools/blob/master/python/postprocessing/modules/btv/btagSFProducer.py#L230), is it not?
The goal of this PR was to change the implementation without changing any behaviour.

Also, at no point a reader object is returned. `reader.eval_auto_bounds()` returns a double (see [BTagCalibrationReader.h](https://github.com/cms-sw/cmssw/blob/aeb38655a73431dbf2eeab4e033d15fbb99cf0df/CondTools/BTau/interface/BTagCalibrationReader.h))

ktht: No, `getSF()` in current implementation *always* returns a float, and the value will be -1 if there's no reader available. The SF is eventually set to 1 if `getSF()` returns too small of a SF:
https://github.com/cms-nanoAOD/nanoAOD-tools/blob/fd69163b283bfcaebe9f6d181db02b41bdf7435a/python/postprocessing/modules/btv/btagSFProducer.py#L277-L280
In the PR you've split up `getSF()` into multiple functions, one part being `getReader()` function which either returns -1 or a reader object ([link](https://github.com/kreczko/nanoAOD-tools/blob/7da623bfebecf3155e5689316fddf7a5c3920d3c/python/postprocessing/modules/btv/btagSFProducer.py#L207-L208)). You get the reader once before looping over systematics ([link](https://github.com/kreczko/nanoAOD-tools/blob/7da623bfebecf3155e5689316fddf7a5c3920d3c/python/postprocessing/modules/btv/btagSFProducer.py#L275)) and then pass it to `getSFs()` in the loop ([link](https://github.com/kreczko/nanoAOD-tools/blob/7da623bfebecf3155e5689316fddf7a5c3920d3c/python/postprocessing/modules/btv/btagSFProducer.py#L278)). Now, the reader could point to either a valid ,,reader'' object if the corrections are available, or to -1 if the corrections are not available. The problem here is that you don't check if the reader passed to `getSFs()` is actually a ,,reader'' object or something else ([here](https://github.com/kreczko/nanoAOD-tools/blob/7da623bfebecf3155e5689316fddf7a5c3920d3c/python/postprocessing/modules/btv/btagSFProducer.py#L241), [here](https://github.com/kreczko/nanoAOD-tools/blob/7da623bfebecf3155e5689316fddf7a5c3920d3c/python/postprocessing/modules/btv/btagSFProducer.py#L243) and [here](https://github.com/kreczko/nanoAOD-tools/blob/7da623bfebecf3155e5689316fddf7a5c3920d3c/python/postprocessing/modules/btv/btagSFProducer.py#L245)). Calling `eval_auto_bounds()` on a float results in a runtime error.

[My solution](https://github.com/HEP-KBFI/nanoAOD-tools/commit/174af42506dcbeed1558a9f4a34b4c40e1d2e998) was to return `None` in `getReader()` if the corrections are missing ([link](https://github.com/HEP-KBFI/nanoAOD-tools/blob/174af42506dcbeed1558a9f4a34b4c40e1d2e998/python/postprocessing/modules/btv/btagSFProducer.py#L207)) so that we won't mix two different data types, and then add a protection to `getSFs()` before calling any functions on the `reader` object ([link](https://github.com/HEP-KBFI/nanoAOD-tools/blob/174af42506dcbeed1558a9f4a34b4c40e1d2e998/python/postprocessing/modules/btv/btagSFProducer.py#L240-L251)).

If it's still confusing, add
```python
btagSF2017_deep = lambda : btagSFProducer("2017", algo = "deepcsv")
```
to your PR and run it.

kreczko: I see. The diff was indeed confusing as there are changes across two functions.
Will make the change.

python/postprocessing/modules/btv/btagSFProducer.py
```diff
@@ -227,6 +227,12 @@ def getFlavorBTV(self, flavor):
         return flavor_btv
 
     def getSFs(self, jet_data, syst, reader,  measurement_type = 'auto', shape_corr = False):
+        if reader is None:
+            if self.verbose > 0
```
ktht: Two problems here: 1) a colon missing missing after `self.verbose > 0`; 2) the SFs are now -1 if the corrections are unavailable (b/c you bypass `sf < 0.01` check in this function), but they should be 1. You can view my fixes here: https://github.com/HEP-KBFI/nanoAOD-tools/commit/692d41170d95ef0c622d180653de455d0f801484

kreczko: Yeah. Unfortunately, that part was done in a hurry. Thanks for checking.

@veelken @rappoccio Are there any plans to add unit tests to nanoAOD-tools?
Since there is a standalone mode, it would be possible to add travic-ci or gilab-ci (like for the CMSSW NanoAOD repo), would it not? I would be happy to write some unit test for this PR if the structure is clear.

arizzi:
@kreczko if you are willing to implement some ci for nanoAOD tools that would be really appreciated.
The plan was to have the mirroring on gitlab and use the ci there... but I just made mirror without implementing anything.
I'd suggest we first write some unit tests then we think about the C.I.

arizzi:
it is not really standalone. most of the modules need cmssw to load some CondFormats or other utility to apply SF.
Using cern gitlab you get cvmfs for free and can do cmsenv to get those libs

kreczko:
I see. Thanks for the explanation, gitlab-ci it is.

BTW: you can have [CVMFS on travis via docker](https://github.com/cms-l1t-offline/cms-l1t-analysis/blob/master/.travis.yml) if you want to, but it will probably be faster at CERN (cached on the CI runner host)

arizzi:
independently of C.I. we can merge this, right?

kreczko:
I think so. I know it works for "normal" cases and the fixes @ktht suggested for the edge-cases are also in.

ktht:
Good to go imo.

PR #53, opened 2018-03-08
fix to haddnano to merge properly Runs tree

scooperstein:


scripts/haddnano.py
```diff
@@ -57,6 +57,7 @@ def zeroFill(tree,brName,brObj) :
 				branchNames.add(br)
 				zeroFill(obj,br,otherObj.GetListOfBranches().FindObject(br))
 			#merge immediately for trees
+                elif isTree:
```
arizzi: this way it doesn't merge Events!!

scooperstein: oops yes of course sorry, should be fixed now

PR #57, opened 2018-03-16
Recompute existing branches

lenzip:
In case one adds a branch that was already there, the existing branch is first scratched from the output tree and the new one is then added.

### DIFF
--- a/python/postprocessing/framework/output.py
+++ b/python/postprocessing/framework/output.py
@@ -12,6 +12,12 @@ class OutputBranch:
         self.buff   = array(_rootBranchType2PythonArray[rootBranchType], n*[0. if rootBranchType in 'FD' else 0])
         self.lenVar = lenVar
         self.n = n
+        #check if a branch was already there. In case SetStatus to 0
+        for b in tree.GetListOfBranches():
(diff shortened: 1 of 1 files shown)


lenzip:
this has a problem. In case more modules are run in the same loop, the second still sees the original branch

lenzip:
I take my previous comment back. This modification actually works even when multiple modules are run in the same loop, provided the modified is accessed from python (when I wrote my earlier comment I was actually using a TTreeFormula, which of course only knows the branches in the input tree). However, uponUpdating the branch, it is propagated to the inputTree with the extraBranch mechanism of InputTree class. Thus when the next module accesses the input tree, if I understand correctly the logic, the modified branch takes precedence on the original one upon calling InputTree.readBranch [(here)](https://github.com/cms-nanoAOD/nanoAOD-tools/blob/master/python/postprocessing/framework/treeReaderArrayTools.py#L51).

lenzip:
Is there anything pending preventing the merge of this pull request?

arizzi:
I guess not, but I did not test it myself. unless @gpetruc or others have objections I'd merge

PR #58, opened 2018-03-16
Option to recompute JECs in postprocessor

arizzi:
This implements the option to redo JEC when computing systematics and also add a producer for data to just reapply the JEC.

redoJEC is ON  by default on 2017 given that we do not have yet NANOAOD with new JECs

@rappoccio @gpetruc @peruzzim


arizzi:
This was tested by VHbb group and seem to work. @rappoccio please have a look

rappoccio:
Sorry for the delay here. This looks reasonable to me. 

One thing we will probably want, though, is a C++ module to speed this up. It may be pretty slow. I haven't gotten around to doing that, so if someone has the time it would be very helpful.

PR #64, opened 2018-03-21
Introduce limited precision for floats in post proc output

arizzi:
Copy reduce mantissa code from CMSSW.
precision can be specified at branch booking time


src/classes.h
```diff
@@ -2,8 +2,10 @@
 #include "PhysicsTools/NanoAODTools/interface/PyJetResolutionScaleFactorWrapper.h"
 #include "PhysicsTools/NanoAODTools/interface/PyJetParametersWrapper.h"
 #include "PhysicsTools/NanoAODTools/interface/WeightCalculatorFromHistogram.h"
+#include "PhysicsTools/NanoAODTools/interface/ReduceMantissa.h"
```
gpetruc: you forgot to git add this file?

arizzi: it seems so

PR #65, opened 2018-03-23
Proposed fix for dropping output branches.

vasile-ghete:
Code changes explained in the corresponding issue.
In addition, added logger to collectionMerger.

### DIFF
--- a/python/postprocessing/framework/output.py
+++ b/python/postprocessing/framework/output.py
@@ -57,13 +57,41 @@ class OutputTree:
         self._file.cd()
         self._tree.Write()
 
+
 class FullOutput(OutputTree):
-    def __init__(self, inputFile, inputTree, outputFile, branchSelection = None, fullClone = False, provenance = False, jsonFilter = None):
+    def __init__(
+            self,
+            inputFile,
+            inputTree,
+            outputFile,
+            branchSelection=None,
+            outputbranchSelection=None,
+            fullClone=False,
+            provenance=False,
+            jsonFilter=None
+    ):
+
         outputFile.cd()
-        if branchSelection: 
-            branchSelection.selectBranches(inputTree)
-        outputTree = inputTree.CopyTree('1') if fullClone else inputTree.CloneTree(0)
+
+        # enable/disable the output branches as requested in outputbranchSelection
+        if outputbranchSelection:
+            outputbranchSelection.selectBranches(inputTree)
+
+        if fullClone:
+            outputTree = inputTree.CopyTree('1')
+        else:
+            
+            outputTree = inputTree.CloneTree(0)
+            
+
+            # enable all branches in inputTree, then disable the branches as requested in branchSelection
+            #
+            inputTree.SetBranchStatus("*",1)
+            if branchSelection:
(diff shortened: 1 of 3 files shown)


arizzi:
it seems you also include a lot of reindent/newlines... and in addition moved a module name to upper case.
Concerning this last point I think all modules are lower case in the postprocessor... do you have any reason for changing that?

vasile-ghete:
I did not change the (python) module name, but the class name. The module is still called collectionMerger.py; the class name was changed from  collectionMerger to CollectionMerger.

This is also a recommendation of PEP-8
https://www.python.org/dev/peps/pep-0008/#class-names
and is now consistent with what is in other modules from post-processor:
https://github.com/cms-nanoAOD/nanoAOD-tools/blob/master/python/postprocessing/framework/branchselection.py
https://github.com/cms-nanoAOD/nanoAOD-tools/blob/master/python/postprocessing/framework/postprocessor.py#L13
etc

Beside the style consistency,  having the class name identical (including capitalization) with the module name can lead in some cases to confusion and errors.

arizzi:
with "module" here I meant Classes inheriting from Module class.
I'm not sure why @gpetruc et al. started with lower case "modules" but I'd try to be consistent... i.e. we can put them all upper case if we think it is better (it may break a few people code but better now than later)

vasile-ghete:
I see now what you mean.

In a somehow contrived example:

import collectionMerger
...
class myCollectionMerger(collectionMerger)

will give an error as myCollectionMerger would inherit from the module, not from the class collectionMerger defined in collectionMerger.py

IMO having class MyModule in module myModule is safer. In practical terms, it makes now with the actual usage no problems, so if you like I can rename the class back to collectionMerger.

gpetruc:
On Fri, Mar 23, 2018 at 4:14 PM, vasile-ghete <notifications@github.com>
wrote:

> I see now what you mean.
>
> In a somehow contrived example:
>
> import collectionMerger
> ...
> class myCollectionMerger(collectionMerger)
>
> will give an error as myCollectionMerger would inherit from the module,
> not from the class collectionMerger defined in collectionMerger.py
>

I don't get the point, if people want to import names from a module they do
"from foo import blah". If they don't, they do "import foo" and then use
"foo.blah".
If they don't know the difference between the two kind of imports, they
have very little chance of being able to write correct python code....

That said, I think most of the classes that I wrote are in UpperCamelCase
(all the framework classes are; and I think I didn't write any of the
modules except a dummy one at the beginning for testing).
So, I'm ok with using UpperCase names for classes.
Module instances (e.g. as in
https://github.com/cms-nanoAOD/nanoAOD-tools/blob/master/python/postprocessing/modules/common/hepmcDump.py#L74
) should be lowerCamelCase.

arizzi:
it is true that people mostly use the instances so a migration would be almost transparent

vasile-ghete:
I close this PR as it mixed more things, a minimal PR for the output branches is created in  https://github.com/cms-nanoAOD/nanoAOD-tools/pull/74

If the other aspects discussed here are thought to be relevant, one can look at them in new issues

Issue #67, opened 2018-03-29
jesUncertaintySources in  jetmetUncertainties.py need to be different for 2016 and 2017


kschweiger:
Running `jetmetUncertainties.py` with __jetmetUncertainties2017All__ crashes because some names of sources are different for 2017 (at least in the referenced files. see [2017](https://raw.githubusercontent.com/cms-nanoAOD/nanoAOD-tools/master/data/jme/Fall17_17Nov2017_V6_MC_UncertaintySources_AK4PFchs.txt) and [2016](https://raw.githubusercontent.com/cms-nanoAOD/nanoAOD-tools/master/data/jme/Summer16_23Sep2016V4_MC_UncertaintySources_AK4PFchs.txt))

Names differing:

| 2016         | 2017       |
|--------------|------------|
| "TimeRunBCD" | "TimeRunB" |
| "TimeRunEF"  | "TimeRunC" |
| "TimeRunG"   | "TimeRunD" |
| "TimeRunH"   | "TimeRunE" |
|              | "TimeRunF" |

I see two possible fixes:
1. Add  a jesUncertaintySources2017 list
2. Since the options are __total__ or __all sources__ it would be a option to get all sources directly from the file instead of hardcoding a list

PR #68, opened 2018-03-30
Proposal for change in jetmetUncertainties.py.

kschweiger:
Addressing #67

Instead of hardcoding the names, they are read from the loaded file


arizzi:
this is now conflicting with some other PRs, can you please rebase/fix the conflict?

kschweiger:
Fixed the conflict.

jpata:
just checking in, are there any issues here?

PR #74, opened 2018-04-20
Minimal PR for dropping output branches.

vasile-ghete:


python/postprocessing/framework/output.py
```diff
@@ -64,11 +64,35 @@ def write(self):
...
+    def __init__(
+            self,
+            inputFile,
+            inputTree,
+            outputFile,
+            branchSelection=None,
+            outputbranchSelection=None,
+            fullClone=False,
+            provenance=False,
+            jsonFilter=None
```
arizzi: any reason for going from one line to multi-line ?

vasile-ghete: very long lines need to be scrolled on github web interface, and are also difficult to read

arizzi: very long lines can be split in fewer lines of reasonable length instead of one line per word... otherwise you now have to scroll vertically instead of having all the code at a glance.

vasile-ghete: for __init_ you see with vertical alignment all arguments at a glance, while a long one-line requires visual parsing (look for comma, look for equal sign, etc).

PR #76, opened 2018-04-20
update btag SF

veelken:
according to today's HN posting from BTV POG: https://hypernews.cern.ch/HyperNews/CMS/get/btag/1525.html


veelken:
I see that Andrea has already updated, so I am closing this PR.

PR #78, opened 2018-04-25
Add puweight sysvar

scooperstein:
PR also includes feature to run JEC uncertainties for AK08 jets without grooming, as is needed to run on 2016 ntuples since GenSubJets are not available in MINIAOD. Sorry I meant for it to be two separate PR's...

Also if anyone knows where I can grab the plus/minus data targets for 2017 I can quickly add that in as well.

@arizzi


scooperstein:
@arizzi now we have pile-up weighting up/down variations for 2016 and 2017. Also adds the protection to run the 2016 AK08 JEC uncertainties without failing due to the lack of the GenSubJet collection. Also I noticed that the wrong file was copied (some html file) for deepCSV V2 .csv fail, so I fixed that.

PR #79, opened 2018-04-25
fix btag sf file

jpata:
Btag scale factor files where not in correct CSV format, fixed that by redownloading from https://twiki.cern.ch/twiki/bin/viewauth/CMS/BtagRecommendation94X

### DIFF
--- a/data/btagSF/CSVv2_94XSF_V2_B_F.csv
+++ b/data/btagSF/CSVv2_94XSF_V2_B_F.csv
@@ -1,488 +1,1918 @@
-
-
-
-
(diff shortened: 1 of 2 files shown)


arizzi:
were they updated? I just did wget...

jpata:
The wget just retrieved the CERN login page html so the code crashes.

arizzi:
ahh ok thx for fixing

Il mer 25 apr 2018, 14:40 Joosep Pata <notifications@github.com> ha scritto:

> The wget just retrieved the CERN login page html so the code crashes.
>

PR #83, opened 2018-05-03
 add protection to muonScaleResProducer.py

scooperstein:
Without this protection in place, some rare events will crash the jobs.


scooperstein:
patch from @HeinerTholen

Issue #81, opened 2018-05-02
overwrite branches with same name


ashrafkasem:
Hi, is there any way to overwrite a branch with the same name 

for example if i have run several modules and i realized that some branches filled incorrectly so i can run again but to fill only those branches instead of running from the beginning. 
i tried but if the branch is already exist in the input file i got error of 
" *** Break *** segmentation violation"

is there any workaround ? 

Cheers
Ashraf

arizzi:
#57 should have introduced this possibility

perhaps @lenzip can help you in case you have problem with it.... @lenzip can we write an example in the workbook https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookNanoAOD ?

lenzip:
My PR actually does not seem to work properly I can only overwite the 
last branch written. I'm working on a fix, I can probably make a PR 
later today.
Giulio

arizzi:
@ashrafkasem  can you try #84 and confirm it works for you too?

lenzip:
The usage is simply that you have to declare the branch as you would 
normally do for a new branch.

arizzi:
@lenzip can you clarify if modules later in the sequence can access the information, and if so how? I think it may not be obvious is modules downstream should still see the original or the updated information

ashrafkasem:
I've tried #84 it looks for me that it works but it does not overwrite the existing branch, it just leave it as it is

lenzip:
That is strange I've tested it multiple times.
Can you give me an example?
Giulio

ashrafkasem:
for examples in the file root://cms-xrd-global.cern.ch//store/user/amohamed/METRun2016Bver1.root
I've filled the branch named "PD_MET" incorrectly and when I rerun it's still unchanged. 
for double check, I filled a new branch with a different name but with the same value and it filled correctly 
looks like the old branch kept the same 

Ashraf

ashrafkasem:
for example the attached module 
[new.py.zip](https://github.com/cms-nanoAOD/nanoAOD-tools/files/1974575/new.py.zip)

ashrafkasem:
okay, it's my mistake. I realized that I was trying to overwrite a branch of type float with boolean so I changed back the type to float and indeed to works. 

sorry for misstating now it works fine with me 

thanks a lot 

Ashraf Mohamed

lenzip:
perfect thanks!

arizzi:
[...]

PR #84, opened 2018-05-03
Recompute existing branches v2

lenzip:
This should enable overwriting branches with the same name, as per issue #81

### DIFF
--- a/python/postprocessing/framework/output.py
+++ b/python/postprocessing/framework/output.py
@@ -12,19 +12,23 @@ class OutputBranch:
         self.buff   = array(_rootBranchType2PythonArray[rootBranchType], n*[0. if rootBranchType in 'FD' else 0])
         self.lenVar = lenVar
         self.n = n
-        #check if a branch was already there. In case, scratch it first 
-        for b in tree.GetListOfBranches():
-          if b.GetName() == name:
-             tree.GetListOfBranches().Remove(b)
-             l= tree.GetLeaf(name)
-             tree.GetListOfLeaves().Remove(l)
-        if lenVar != None:
+        #check if a branch was already there 
+        existingBranch = tree.GetBranch(name)
+        print name, existingBranch
+        if (existingBranch):
+          #print name,"is an existing branch"
+          self.branch = existingBranch
+          self.branch.SetAddress(self.buff)
+        else:  
+          #print name,"is a new branch"
+          if lenVar != None:
             self.branch = tree.Branch(name, self.buff, "%s[%s]/%s" % (name,lenVar,rootBranchType))
-        elif n == 1:
+          elif n == 1:
             self.branch = tree.Branch(name, self.buff, name+"/"+rootBranchType)
-        else:
+          else:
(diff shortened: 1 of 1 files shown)


arizzi:
would this allow to read the value in the old tree?

arizzi:
ah this is just the output branch, so I guess it works...

python/postprocessing/framework/output.py
```diff
@@ -44,7 +45,7 @@ def __init__(self, tfile, ttree, intree):
         self._intree = intree
         self._branches = {} 
     def branch(self, name, rootBranchType, n=1, lenVar=None, title=None):
-        if (lenVar != None) and (lenVar not in self._branches) and (not self._tree.GetBranch(lenVar)):
+        if (lenVar != None) and (lenVar not in self._branches): #and (not self._tree.GetBranch(lenVar)):
```
arizzi: @lenzip @gpetruc  why the last AND was needed / not needed anymore ?

lenzip:
Hi Andrea
     the last "and" was making it impossible to overwrite the length var.
I'm unsure about the original need.
Giulio

arizzi:
and why do we want to overwrite the len var? when we rewrite it, are we
sure that  all branches using that lenvar are of the correct length?

lenzip:
For example, if we want to remove the leptons from the jets collection, 
the length of the jet collection changes.
In such a case one needs to make sure that all Jet_* branches are 
recomputed.
One example is here:
declaration of the branches: 
https://github.com/latinos/LatinoAnalysis/blob/master/NanoGardener/python/modules/LeptonSelLazy.py#L96-L98
writing of the branches: 
https://github.com/latinos/LatinoAnalysis/blob/master/NanoGardener/python/modules/LeptonSelLazy.py#L327-L334

arizzi:
ok, but can we enforce that all collection are updated ?

arizzi:
btw this also breaks the indices, so I think pruning collections (with same
name) should not be supported...

lenzip:
On 08/05/2018 10:43, arizzi wrote:
> btw this also breaks the indices, so I think pruning collections (with 
> same
> name) should not be supported...
>
I'm afraid you are correct. I had not thought of this.
Should maybe some king of light of light cloning (e.g. just store the 
indices of the original objects in the pruned collection) be implemented?

arizzi:
i.e. producing something like
selJetIdxes[newLen]

then accessing
Events->Draw("Jet_pt[selJetIdxes]")

xjanssen:
Well we are already doing a custom made Lepton_XXX with XXX = pt, eta, phi , charge, type , iDx to Muon_ or Electron_ -> so we can avoid cleaning the original collections to be safe and we will do the same for Jets_ --> CleanJets_, still we will need to overwrite the   CleanJets_ and Lepton_ collections at some points but these would be our own collections, not the miniAOD ones.

arizzi:
we could probably force some checks such that if you change the length then we automatically
- drop from output all the elements of the original collection
- drop from output all indices pointing there (the naming schema should allow to find that)

PR #85, opened 2018-05-04
WIP: nanoAOD tree name configurable

jpata:
Don't merge yet!
To make it possible to run the postprocessor on nanoAOD trees in case the Events tree was renamed, moved etc. Shouldn't affect otherwise.


jpata:
Actually, this won't work with `FullOutput`, so we rather adapt in a different place. Nevermind.

PR #90, opened 2018-05-23
JER files for 2017

ktht:


arizzi:
are they really JER 2017 or rather the 2016 final?

ktht:
I have to admit that I got confused as well for a moment, b/c the Fall17 files were pushed w/ the same commit as Summer16 files. So I asked on HN (see the references in my last commit message).

A somewhat related question: why not require JRDatabase to be a dependency of this repository or add it as a git submodule? Or at the very least read from SQLite DB/tarballs to avoid unpacking the tarballs? It would be more convenient to submit PRs when JERC drops a new set of SFs.

PR #94, opened 2018-06-08
Updating JECs, and adding an HT module

rappoccio:
The JECs for the 2016 data were not present. If we want to add the recalibrator, we should add them.

I also added an HT module, seems useful. 

(EDIT: I meant 2016 data)


arizzi:
I see already some 
Fall17_17Nov2017B_V6_DATA 
aren't these more up to date than the one you posted?

concerning HT I think we had an existing module from day zero that included also lepton cleaning

arizzi:
ah sorry it was an MHT calculator

arizzi:
is this still relevant or can it be closed?

rappoccio:
This is still needed for the HT producer. @camclean are the txt files obsolete or are they still needed here?

camclean:
The 2017 text files are up to date. The 2016 text files are the most recent ones not affected by the formula evaluator bug, so I think we should keep them. 

Next, I can add the 2018 and most recent, but buggy, 2016 text files. Though I think that should be in a later PR.

fgolf:
@rappoccio 

Sorry this has languished, getting back to it now.  The JEC updates were merged as part of PR #133, but it would be good to add the HT module.

As there have been a number of changes since this PR was made, only 3 of the files are needed so I thought it would be easier to make a new PR, #157 does this.  Ran basic tests and everything looks as expected.  If you're ok with this, I'll go ahead and merge it.

Thanks and sorry again for this sitting for so long.

fgolf:
closing, superseded by #157

PR #99, opened 2018-08-23
Nominal correction factors in jetmetUncertainties

kschweiger:
This PR adds branches for the nominal correction factors (JEC and JER) applied within the jetmetUncertainties post processing module. This is necessary to recalculate the uncorrected pt after post processing. W/o this change it is not possible to do this if a new JEC is applied in module.

**New Branches:**
- Jet_corr_JEC *(is equal to 1/(1-rawFactor) if no new JEC is applied in module)*
- Jet_corr_JER

Defined so that: Jet_pt_nom = Jet_pt_raw * Jet_corr_JEC * Jet_corr_JER


kschweiger:
It been some time since posting this. Is there a problem with this PR?

arizzi:
I guess only summer and switch of XPOG conveners that is happening now.
Anyhow another option would be to recalulate the new raw-factor, isn't it?

anyhow unless other object ~today I will merge later

Issue #106, opened 2018-10-04
Function InputTree does not work without an entrylist


lcorcodilos:
`InputTree` function in postprocessing/framework/treeReaderArrayTools.py does not account for `entrylist == None` when calling `TTreeReader`.

I'll submit a pull request to account for this and call the appropriate initializer for `TTreeReader`.

lcorcodilos:
Submitted pull request #107

lcorcodilos:
Solution can be found in the pull request

PR #107, opened 2018-10-04
InputTree can now accomodate entrylist=None

lcorcodilos:
Without this, an entrylist is required in the call to InputTree.

Fix for issue #106.

### DIFF
--- a/python/postprocessing/framework/treeReaderArrayTools.py
+++ b/python/postprocessing/framework/treeReaderArrayTools.py
@@ -7,7 +7,10 @@ def InputTree(tree,entrylist=None):
     if hasattr(tree, '_ttreereader'): return tree # don't initialize twice
     tree.entry = -1
     tree._entrylist = entrylist
-    tree._ttreereader = ROOT.TTreeReader(tree,tree._entrylist)
+    if tree._entrylist == None:
+        tree._ttreereader = ROOT.TTreeReader(tree)
+    else:
+        tree._ttreereader = ROOT.TTreeReader(tree,tree._entrylist)
     tree._ttreereader._isClean = True
     tree._ttrvs = {}
     tree._ttras = {}


lcorcodilos:
So it appears the solution is not as simple as the one I've provided since `_remakeAllReaders` has the same issue. Sorry for jumping the gun! I'll keep looking into this though I imagine it might just be simplest to create an un-skimmed `entrylist`via `preSkim`. However at a minimum, the definition for `InputTree` should not initialize `entrylist = None`if the function cannot accommodate it.

lcorcodilos:
So it appears I can't run this with anything like the below code:

`f = TFile.Open('BstarLH2200_jetNano.root')`
`inTree = f.Get("Events")`
`elist,jsonFilter = preSkim(inTree, None, '')`
`inTree= InputTree(inTree,elist)`

I get 

```
File "/uscms_data/d3/lcorcodi/BStar13TeV/CMSSW_8_1_0/python/PhysicsTools/NanoAODTools/postprocessing/framework/treeReaderArrayTools.py", line 13, in InputTree
    tree._ttreereader = ROOT.TTreeReader(tree,tree._entrylist)
TypeError: none of the 5 overloaded methods succeeded. Full details:
  TTreeReader::TTreeReader() =>
    takes at most 0 arguments (2 given)
  TTreeReader::TTreeReader(TTree* tree) =>
    takes at most 1 arguments (2 given)
  TTreeReader::TTreeReader(const char* keyname, TDirectory* dir = __null) =>
    could not convert argument 1 (expected string or Unicode object, TTree found)
  TTreeReader::TTreeReader(const char*, TFileCollection*) =>
    could not convert argument 1 (expected string or Unicode object, TTree found)
  TTreeReader::TTreeReader(const TTreeReader&) =>
    takes at most 1 arguments (2 given)
```

I'm in CMSSW_8_1_0 with ROOT version 6.06/09.

Maybe this is more appropriate to discuss under issues but could someone please advise?

lcorcodilos:
So I must apologize. The TTreeReader constructor required by `InputTree` exists but it's in a newer versions of ROOT than what I was using. I'll close the pull request and update the issue entry.

Issue #108, opened 2018-10-25
LeptonEfficiencyCorrector not found when running with crab


amagnan:
Dear experts,
I have been running the leptonSFproducer - which calls the c++ code in LeptonEfficiencyCorrector - properly interactively, but when I submit to crab I get this error.
Could you please indicate me what is needed such that this file would be found by crab ?
Many thanks.

== CMSSW: /srv/CMSSW_9_4_10
== CMSSW:
== CMSSW: /srv
== CMSSW: mv: cannot stat `module': No such file or directory
== CMSSW: Found Proxy in: /srv/aebf959e130b2b40234b483b07243972d2e6d8eb
== CMSSW: Error in <PyROOT::TPyROOTApplication::ProcessLine>: macro /srv/CMSSW_9_4_10/src/PhysicsTools/NanoAODTools/python/postprocessing/helpers/LeptonEfficiencyCorrector.cc not found in path .:/cvmfs/cms.cern.ch/slc6_amd64_gcc630/lcg/root/6.10.08-elfike2/macros
== CMSSW: Error in <PyROOT::TPyROOTApplication::ProcessLine>: macro /srv/CMSSW_9_4_10/src/PhysicsTools/NanoAODTools/python/postprocessing/helpers/LeptonEfficiencyCorrector.cc not found in path .:/cvmfs/cms.cern.ch/slc6_amd64_gcc630/lcg/root/6.10.08-elfike2/macros
== CMSSW: ARGV: ['crab_script_vbf_mc.py', '2']
== CMSSW: cms.untracked.vstring('/store/mc/RunIIFall17NanoAOD/DYJetsToLL_1J_TuneCP5_13TeV-amcatnloFXFX-pythia8/NANOAODSIM/PU2017_12Apr2018_94X_mc2017_realistic_v14-v1/00000/0AC4DA82-D591-E811-BE97-00266CFFCD50.root')
== CMSSW: --------------- using edmFileUtil to convert PFN to LFN -------------------------
== CMSSW: Data is local
[...]

PR #110, opened 2018-11-26
Moving leptonSF code to compiled class

amagnan:
- the previous version, with a helper macro, was working fine interactively, but not in crab. Makes more sense to have properly compiled CppWorker. 
- Added possibility to produce loose or tight working points in the leptonSFProducer.py  -> POG still needs to extend to other WP, and compute errors properly -> need source-by-source variations, how it is done now is incorrect I believe (but not saved in output branches at the moment anyway).


amagnan:
Related to issue https://github.com/cms-nanoAOD/nanoAOD-tools/issues/108

arizzi:
can you fix the conflicts so we may consider for merge? (apologies for the late feedback)

amagnan:
I fixed the merge error (removing the python helper code which was moved in proper compiled cpp class) and also I fixed the SF error to be returned by "type", up to the user then to combine properly the different errors in quadrature or keep as separate ....

arizzi:
@sroychow can you validate / cross check with the similar code you have?

sroychow:
@arizzi My code was updated only for Muons. I will check if the same will work for electrons and get back to you.

sroychow:
@arizzi  
Please have a look at 
https://github.com/sroychow/nanoAOD-tools/blob/a1cefbe44f496e4c95ce6e8157383f13c61a9311/python/postprocessing/modules/common/lepSFProducer_v2.py

The module takes as input histograms(depending on working points) to read SF and uncertainties from the MuonPOG trees. It only uses the WeightCalculater helper class. I have also checked the root files  provided for electrons and the module can be used for egamma case as well.

python/postprocessing/modules/common/lepSFProducer.py
```diff
@@ -26,18 +55,20 @@ def __init__(self, muonSelectionTag, electronSelectionTag):
         self.el_h = ROOT.std.vector(str)(len(el_f))
         for i in range(len(el_f)): self.el_f[i] = el_f[i]; self.el_h[i] = el_h[i];
 
-        if "/LeptonEfficiencyCorrector_cc.so" not in ROOT.gSystem.GetLibraries():
-            print "Load C++ Worker"
-            ROOT.gROOT.ProcessLine(".L %s/src/PhysicsTools/NanoAODTools/python/postprocessing/helpers/LeptonEfficiencyCorrector.cc+" % os.environ['CMSSW_BASE'])
+        for library in [ "libCondFormatsJetMETObjects", "libPhysicsToolsNanoAODTools" ]:
```
arizzi: why are we checking for METObjects?

PR #113, opened 2018-12-21
Updated to V32 JEC in jetmetUncertainties.py

kschweiger:
Added necessary JEC files from https://github.com/cms-jet/JECDatabase to `data` and updated names in postprocessor module

Issue #98, opened 2018-07-19
JetMetuncertainty for 2017 


ashrafkasem:
Hi All, 
I've tried to get the jetmetuncertainty for 2017 dataset and I've used the default module in modules/jme using "jetmetUncertainties2017All" option but every time something is broken and I got the message below. 

any ideas? 
Loading jet energy scale (JES) uncertainties from file PhysicsTools/NanoAODTools/data/jme/Fall17_17Nov2017_V6_MC_Uncertainty_AK4PFchs.txt'
%MSG-e JetCorrectorParameters:
No definitions found!!!
%MSG

 *** Break *** segmentation violation

lcorcodilos:
Hi @ashrafkasem,

I've been spending the day writing a PR to update all of the JEC calculations and just came across the same error as you have above. Did you ever solve this?

Thanks!

arizzi:
@bianchini has a fix IIRC

lcorcodilos:
I "fixed" it by changing [this line](https://github.com/lcorcodilos/nanoAOD-tools/blob/046cf7a1969ecbf05ec81b76d3cdfa918aaffea4/python/postprocessing/modules/jme/jetmetUncertainties.py#L106) to 

`if (self.era == "2016" or self.era == "2017") and jesUncertainty == 'Total' and len(self.jesUncertainties) == 1:` 

But I'm not so sure of the significance of jesUncertainty_label. I'm assuming that some of the JEC text files have a label per set of values and that was absent in only 2016 but is now absent in 2017 as well - that's just my guess based on looking at the code for a few minutes though :-)

PR #126, opened 2019-03-19
Update JECs and scripts for 2017 and preliminary 2018

lcorcodilos:
Hi all,

Based on my reading of https://twiki.cern.ch/twiki/bin/view/CMS/JECDataMC, it seems that the jetmetUncertainties.py and jetRecalib.py modules are out of date. This PR should ammend the scripts and add the new JEC/R files to /data/jme/. I think there's an argument for removing the old txt files to avoid confusion and keep the repo cleaner but there may be a reason to keep them that I haven't thought so I've left them for now.

This is my first time having to handle these so please double check my work :-)

In particular, I'm not sure why redoJEC=True for jetmetUncertainties2017All. I've set it to False for 2018 but that was just a guess.

Additionally, I came across an issue that noGroom=True does not work for any year. All of the msoftdrop branches generated in jetmetUncertainties.py are filled with -1. Through some print outs, I traced this back to the [genJets and subGenJets collections](https://github.com/cms-nanoAOD/nanoAOD-tools/blob/master/python/postprocessing/modules/jme/jetmetUncertainties.py#L161-L162) being empty. I downloaded a central NanoAOD file to confirm that this is NOT true and tested to see that this is not unique to the PR code. I didn't dig into this further since it's clear that my code didn't cause (or solve) the issue but I figure it's worth mentioning. If someone has a fix, I can implement it in this PR.

Lucas


lcorcodilos:
Some of my changes are identical to what was proposed in #113 but mine include 2018 values, a fix to issue #98, and updates to jetRecalib.py

arizzi:
@kschweiger @ashrafkasem can you cross check and comment?

lcorcodilos:
Adding @pmaksim1

lcorcodilos:
Also adding @rappoccio @UBParker. I'd add Christine as well but I can't find her

lcorcodilos:
Sorry - clicked the wrong button!

arizzi:
I think there are way too many files pulled in by this PR, do we really need JECs for each separate run period? I guess not.
I would also suggest that for better handling of so many conditions we create a tgz file for each relevant GT and we unpack it on the fly when running, or  something like it.

Can you separate the fixes in the py code from the update of the txt files in two PRs so that the code is actually reviwable ?

rappoccio:
> I think there are way too many files pulled in by this PR, do we really need JECs for each separate run period? I guess not.

We do, unfortunately. This complication is the primary reason I'm not thrilled about the re-application of JECs at the NANOAOD level. It seems to be popular instead of just remaking them as soon as new JECs come out, so here we are. 

> I would also suggest that for better handling of so many conditions we create a tgz file for each relevant GT and we unpack it on the fly when running, or something like it.
> 

That may be a wiser choice, I agree. 

> Can you separate the fixes in the py code from the update of the txt files in two PRs so that the code is actually reviwable ?

Also +1 for this.

lcorcodilos:
I also just noticed that a commit from October sneaked in here.

So here's my plan then. I'll separate my fork into three branches and revert them all back to the original fork state. Then each fork will have a single commit 
- One for the python module changes. I'll put something in the module to untar in `__init__` and delete the untarred folder in `endJob`(?)
- One for the JECs which I'll tar together 
- One for my stray commit to treeReaderArrayTools.py

Then we can close this PR and start from those.

lcorcodilos:
I have a follow up coding question to this. I was looking into good ways to access tgz contents and came across `tarfile` in python. 

We can do something like 
`import tarfile`
`self.tar = tarfile.open(globalTag + '.tgz', 'r:gz')`
`with self.tar.extractfile(self.jesUncertaintyInputFileName) as f:`
`    ....`

and now `f` can be used as normal but we never extract the txt file permanently in any directory (so there's no cleanup). Sound reasonable?

lcorcodilos:
It's occurred to me that this won't work with `JetCorrectorParameters` because that requires a string input so I'll instead use `tarfile.extractall()` to a get everything in a scratch directory that gets tossed at the end.

lcorcodilos:
New PRs created. I think it's safe to close this one now.

lcorcodilos:
Just wanted to leave a comment about the "noGroom=True does not work for any year" issue I originally commented on. The fault was on my end - one needs to make sure that GenJets is "kept" in the keep and drop file so that the module can access it.

PR #127, opened 2019-03-21
fix bug in filling weighted count histogram

AndrewLevin:
The weighted histogram should be filled with a negative weight in the same bin.

### DIFF
--- a/python/postprocessing/modules/common/countHistogramsModule.py
+++ b/python/postprocessing/modules/common/countHistogramsModule.py
@@ -11,22 +11,22 @@ class countHistogramsProducer(Module):
     def endJob(self):
         pass
     def beginFile(self, inputFile, outputFile, inputTree, wrappedOutputTree):
-        self.h_nevents=ROOT.TH1F('nEvents',   'nEvents',   1, 0, 1)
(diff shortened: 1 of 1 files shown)


arizzi:
well it is commented out code and it really depends on what you want to count... counting individually positive and negatives may have adavantages

AndrewLevin:
I just pushed a commit which uncomments the filling of the weighted histogram and automatically detects when there is a generator weight. I think it makes more sense to fill the positive and negative weights in a single bin, since the sum of them is what people typically want, and you can recover the separate numbers by combining with the unweighted number of events histogram.

arizzi:
I think changing something that other people may already be using would need a stronger reason.
Given that all representations allow to recover one or another number, I would prefer not to change the meaning.

AndrewLevin:
Well, currently, it fills a histogram with a range from 0 to 1 at -0.5. I think it is probably more likely that someone is using it incorrectly now i.e. using the bin 0 to 1 as if it had both positive and negative weights.

arizzi:
ok let's change the name then to "nEventsGenWeighted"

arizzi:
I mean so that if anyone was using the old one we do not break the interface

AndrewLevin:
ok, how about this? I changed the name to `nEventsGenWeighted` and added back the old histogram commented like it was before.

PR #130, opened 2019-04-01
Secondary file input to postprocessor

jhgoh:
This PR suggests to modify postprocessor to support friend tree in the input.

Use case: Users have centrally produced NanoAOD root file and Friend tree. 
They want to read branches across two (or >2) files.

Usage:
```
nano_postproc nano_postproc.py --friend -I Some.User.Module userModule nanoAOD.root,friend.root,friend2.root
```

### DIFF
--- a/python/postprocessing/framework/postprocessor.py
+++ b/python/postprocessing/framework/postprocessor.py
@@ -80,6 +80,10 @@ class PostProcessor :
         t0 = time.clock()
 	totEntriesRead=0
 	for fname in self.inputFiles:
+	    ffnames = []
+	    if "," in fname:
+	        fnames = fname.split(',')
+	        fname, ffnames = fnames[0], fnames[1:]
 
 	    # open input file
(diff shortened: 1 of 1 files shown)


arizzi:
nice feature, did you explicitly check that both the input branch selection and the output branch selection works with the friend trees?

jhgoh:
@arizzi yes, I tested with simple postprocessor which prints out variables from two different root files. (branch names are arbitrary in the code snippet below)

```python
   # ... code snippet from my postprocessor module
    def analyze(self, event):
        """process event, return True (go to next module) or False (fail, go to next event)"""
        print "MET_pt", event.MET_pt, "RivetJet_pt[0]", event.RivetJet_pt[0]
```

prints out values what I expect from each input root files in the same row.

arizzi:
it would be probably nice to add event,rum,lumi on the Friends and explicitly AddFriend with an index created on those to enable also skimmed friends

PR #132, opened 2019-04-02
JME module update for 2017 and (preliminary) 2018 JECs

lcorcodilos:
Module updates separated out from PR #126.

Accommodates the fact that JECs will be stored in tgz files now.

### DIFF
--- a/python/postprocessing/modules/jme/jetRecalib.py
+++ b/python/postprocessing/modules/jme/jetRecalib.py
@@ -103,11 +103,11 @@ jetRecalib2017DEAK8Puppi = lambda : jetRecalib("Fall17_17Nov2017DE_V32_DATA",jet
 jetRecalib2017FAK8Puppi = lambda : jetRecalib("Fall17_17Nov2017F_V32_DATA",jetType="AK8PFPuppi")
 
 jetRecalib2018A = lambda : jetRecalib("Autumn18_RunA_V8_DATA")
-jetRecalib2018B = lambda : jetRecalib("Autumn18_RunA_V8_DATA")
-jetRecalib2018C = lambda : jetRecalib("Autumn18_RunA_V8_DATA")
-jetRecalib2018D = lambda : jetRecalib("Autumn18_RunA_V8_DATA")
+jetRecalib2018B = lambda : jetRecalib("Autumn18_RunB_V8_DATA")
+jetRecalib2018C = lambda : jetRecalib("Autumn18_RunC_V8_DATA")
+jetRecalib2018D = lambda : jetRecalib("Autumn18_RunD_V8_DATA")
 
 jetRecalib2018AAK8Puppi = lambda : jetRecalib("Autumn18_RunA_V8_DATA",jetType="AK8PFPuppi")
-jetRecalib2018BAK8Puppi = lambda : jetRecalib("Autumn18_RunA_V8_DATA",jetType="AK8PFPuppi")
-jetRecalib2018CAK8Puppi = lambda : jetRecalib("Autumn18_RunA_V8_DATA",jetType="AK8PFPuppi")
-jetRecalib2018DAK8Puppi = lambda : jetRecalib("Autumn18_RunA_V8_DATA",jetType="AK8PFPuppi")
+jetRecalib2018BAK8Puppi = lambda : jetRecalib("Autumn18_RunB_V8_DATA",jetType="AK8PFPuppi")
+jetRecalib2018CAK8Puppi = lambda : jetRecalib("Autumn18_RunC_V8_DATA",jetType="AK8PFPuppi")
+jetRecalib2018DAK8Puppi = lambda : jetRecalib("Autumn18_RunD_V8_DATA",jetType="AK8PFPuppi")
--- a/python/postprocessing/modules/jme/jetRecalib.py
+++ b/python/postprocessing/modules/jme/jetRecalib.py
@@ -24,7 +24,7 @@ class jetRecalib(Module):
         self.jmsVals = [1.00, 0.99, 1.01]
         
 
-        self.jesInputFilePath = os.environ['CMSSW_BASE'] + "/src/PhysicsTools/NanoAODTools/data/jme/"
+        self.jesInputFilePath = os.environ['CMSSW_BASE'] + "/src/PhysicsTools/NanoAODTools/data/jme/" 
 
         self.jetReCalibrator = JetReCalibrator(globalTag, jetType , True, self.jesInputFilePath, calculateSeparateCorrections = False, calculateType1METCorrection  = False)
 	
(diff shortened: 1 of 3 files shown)


lcorcodilos:
@arizzi @rappoccio @pmaksim1

python/postprocessing/modules/jme/jetmetUncertainties.py
```diff
@@ -23,8 +23,13 @@ def __init__(self, era, globalTag, jesUncertainties = [ "Total" ], jetType = "AK
         self.jesUncertainties = jesUncertainties
 
         # smear jet pT to account for measured difference in JER between data and simulation.
-        self.jerInputFileName = "Spring16_25nsV10_MC_PtResolution_" + jetType + ".txt"
-        self.jerUncertaintyInputFileName = "Spring16_25nsV10_MC_SF_" + jetType + ".txt"
+        if era == "2016":
+            self.jerInputFileName = "Summer16_25nsV1_MC_PtResolution_" + jetType + ".txt"
+            self.jerUncertaintyInputFileName = "Summer16_25nsV1_MC_SF_" + jetType + ".txt"
```
arizzi: isn't it possible to detect those from the GT names or at least to have those not hardcoded here?

arizzi:
can anyhow run a quick validation that we obtain the same results with this and previous implementation and post a few plots for reference?

python/postprocessing/modules/jme/jetmetUncertainties.py
lcorcodilos: According to [L18-L21](https://github.com/cms-nanoAOD/nanoAOD-tools/blob/1995a8054ffa8738e34055668fb308d92a2f89f7/python/postprocessing/modules/jme/jetmetUncertainties.py#L18-L21) we can't derive these straight from the GT because the JES and JER global tags don't match (I've confirmed this is still true when I looked up the recommendations). So JES name comes from the global tag and the JERs are hardcoded to year.

We could hardcode a dictionary which does a JER tag lookup based on the global tag or we could build in another argument in the class for just the JER tag. Or maybe a combination of the two? Have a hard coded dictionary lookup by default but it's also possible to provide an argument for the JER tag that allows the user to force the tag without needing to change the code.

lcorcodilos:
Working on the validation now. I'm running test ttbar MC and data files for 2016, 2017, and 2018. I think that covers all scenarios.

python/postprocessing/modules/jme/jetSmearer.py
```diff
@@ -17,7 +17,12 @@ def __init__(self, globalTag, jetType = "AK4PFchs", jerInputFileName = "Spring16
 
         # read jet energy resolution (JER) and JER scale factors and uncertainties
         # (the txt files were downloaded from https://github.com/cms-jet/JRDatabase/tree/master/textFiles/Spring16_25nsV10_MC )
-        self.jerInputFilePath = os.environ['CMSSW_BASE'] + "/src/PhysicsTools/NanoAODTools/data/jme/"
+        # Text files are now tarred so must extract first
+        self.jerInputArchivePath = os.environ['CMSSW_BASE'] + "/src/PhysicsTools/NanoAODTools/data/jme/"
+        self.jerTag = jerInputFileName[:jerInputFileName.find('_MC_')+len('_MC')]
+        self.jerArchive = tarfile.open(self.jerInputArchivePath+self.jerTag+".tgz", "r:gz")
+        self.jerInputFilePath = "scratch/"+self.jerTag
```
arizzi: we should rather create a unique name with some mktemp function to avoid clashes

python/postprocessing/modules/jme/jetmetUncertainties.py
```diff
@@ -103,15 +117,15 @@ def beginJob(self):
...
+            if jesUncertainty == 'Total' and len(self.jesUncertainties) == 1:
                 jesUncertainty_label = ''
             pars = ROOT.JetCorrectorParameters(os.path.join(self.jesInputFilePath, self.jesUncertaintyInputFileName),jesUncertainty_label)
             self.jesUncertainty[jesUncertainty] = ROOT.JetCorrectionUncertainty(pars)    
 
         self.jetSmearer.beginJob()
 
     def endJob(self):
-
+        shutil.rmtree("scratch/")
```
arizzi: this is dangerous, let's use a proper temp file

python/postprocessing/modules/jme/jetSmearer.py
lcorcodilos: I did not know mktemp was a thing! It makes sense of course, I just never thought to look for a dedicated tool :-) I've been playing with the built-in version for python from `tempfile` and it works as intended in my minimum working example. It doesn't look like we need to remove the directory ourselves (it gets deleted once python cleans up the object) so I'll remove the call to shutil as well. Will commit ASAP.

lcorcodilos:
Okay, please confirm that the new temporary directory implementation is satisfactory. I also fixed a copy-paste error.

cmantill:
Hi all, I am also implementing the latest JECs and JER for the legacy dataset and came upon a doubt which may also concern this implementation in nanoAOD. For 2017/18 JER I find some of the unc. to have values < 1 e.g.: https://github.com/cms-jet/JRDatabase/blob/master/textFiles/Fall17_V3_MC/Fall17_V3_MC_SF_AK8PFPuppi.txt#L18. `SFUp=0.974`.

But this produces a nan value when trying to propagate into up and down Uncertainties e.g. in this line: https://github.com/cms-nanoAOD/nanoAOD-tools/blob/763e409d4d71d0d3379fa262437f512d7a4c82a1/python/postprocessing/modules/jme/jetSmearer.py#L115 (sqrt(sf-1)= sqrt(negative value)). I have provisionally taken the abs value since I expect a minimal effect on the analysis but is there a better way to consider this case? 

I will try to forward this comment to those in charge of deriving the corrections but if you have any other feedback please let me know since I imagine it will also happen for nanoAOD (unless I have a bug somewhere in my implementation).

arizzi:
clearly we cannot unsmear when we have no matched gen, I'd say the safest thing is to do nothing in that case

adewit:
Hi @lcorcodilos, I used this PR and the corresponding one with the tar archives to add updated JES uncertainties / re-apply the corrections for 2018 to some samples I'm postprocessing. To make everything work for data and simulation I had to make a small number of changes, which - in case it is useful going forward - you can find here: https://github.com/adewit/nanoAOD-tools/commit/30781231a14d06590f7e3f1faa060590bbfee10f

sroychow:
Hi @arizzi @lcorcodilos I did a test with this PR and seems to work fine. The updates from @adewit  should also be included so that the reading of tar files are consistent in jetRecalib.py and jetmetUncertainties.py.

arizzi:
@lcorcodilos <https://github.com/lcorcodilos> can you include them?

lcorcodilos:
Hi all,

@cmantill, it seems @arizzi's suggestion is already implemented in the code so I'm leaving this issue alone.

@adewit and @sroychow, thank you both for checking out and testing the code! I had to step away from this for several days and didn't have a chance to respond earlier. I've implemented the changes in the latest commit.

So that I understand, could someone explain or point me to an explanation that describes when the redoJEC flag should be set to true?

python/postprocessing/modules/jme/jetRecalib.py
```diff
@@ -24,7 +24,11 @@ def __init__(self,  globalTag, jetType = "AK4PFchs"):
         self.jmsVals = [1.00, 0.99, 1.01]
```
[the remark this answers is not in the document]
lcorcodilos: Removed

PR #134, opened 2019-04-11
Muon rochester Run2 V3 added

sroychow:
Updates to Muon Rochester correction module.


fgolf:
Before commenting on specifics, I have some general questions about the structure of this PR.  Is there a reason this code is packaged with the data and the code that reads and applies the corrections changes with each version of the corrections?  I think it would be preferable to restructure this, for example, move RoccoR.h/cc into helpers and create a producer in a new sub-directory modules/muon that could be then used similar to other modules in the framework.

arizzi:
@fgolf rochester corrections are provided this way. I think this request of decoupling code from data should be passed to Muon POG and the guys providing the corrections, it has little to do with this PR

fgolf:
@arizzi OK, I will follow up with the Muon POG.

python/postprocessing/data/roccor.Run2.v3/RoccoR.cc
```diff
@@ -0,0 +1,337 @@
...
+    int H = etaBin(fabs(eta));
+    int F = trkBin(v, H); 
+    const ResParams &rp = resol[H];
+    double x = rp.kRes[type] * Sigma(pt, H, F) * rp.cb[F].invcdf(u);
+    return 1.0/(1.0+x);
+}
+
+double RocRes::kSmear(double pt, double eta, TYPE type, double w, double u, int n) const{
+    int H = etaBin(fabs(eta));
+    int F = n-NMIN;
```
fgolf: Should this line also be changed to: 
int F = n>NMIN ? n-NMIN : 0; ?

sroychow: Change has been implemented.

PR #149, opened 2019-05-24
Correct JEC implementation

camclean:
Corrected the order of redo jec and jer smearing, apply jec to both mass and pt, and have raw mass/pt branches.

@arizzi let me know if there's anything else you want me to do before going ahead with these changes.


camclean:
@arizzi are there any problems merging this pull request? This is important for those who are redoing JECs. @clelange @lathomas

gmandorl:
@camclean  I tried to apply correction with the new files and with the old files on the same sample and I obtained identical Jet_pt_nom and Jet_pt_mass

gmandorl:
@camclean I am sorry I ment Jet_mass_nom

camclean:
Can you be more specific with what you did? You found that jet_pt_nom = jet_mass_nom? Or just that jet_pt_nom didn't change? Which files and which corrections did you use? Did you turn on redoJEC?

gmandorl:
@camclean I realized I was using jetmetUncertainties2017 on 2017 sample, that is why jet_pt_nom didn't change. I just tried using jetmetUncertainties2018 and I see differences in jet_pt_nom and jet_mass_nom.
Sorry for my mistake

arizzi:
@camclean wasn't the claim in the other PR that even rerunnin on same JEC as nanoaod you would expect pt_nom to be wrong? Is what @gmandorl sees for 2017 compatible with your expectations?

camclean:
No to your first question. You asked "if the JEC one applies in postprocessing are the same already applied in the production of nanoaod, is the output of the current module correct (i.e. the JEC/JER order doesn't matter)?" To which I responded, "That is correct. You can check this by comparing FatJet_pt_nom/FatJet_corr_JER and FatJet_pt, which should give you the same numbers." I did edit my answer, which was at first incorrect. Apologies for the confusion.

What @gmandorl sees is compatible with my expectations because the latest version of the 2017 nanoAOD has the most recent corrections applied. Therefore, recorrecting with the same corrections will not make a difference.

The difference he saw with the 2018 pt_nom and mass_nom (between the old and new versions of the code) was the fact that:
a.) the jer_nom_val (FatJet_pt_corr_JER in the tree) was derived with different JEC versions applied.
b.) the mass was only recorrected (with JECs) in the new version. In the old version, the mass would still have the old JECs applied.

Issue #158, opened 2019-06-07
Add raw and jet correction branches to data (jetRecalib.py)


camclean:
Jet correction factor and raw mass, pt branches should be added to the data trees to make them consistent with MC. Then an option to redo JECs should be added so that the new branches can be added without redoing the corrections.

Issue #155, opened 2019-05-29
In jetmetUncertainties.py, the 2016 JES uncertainty files need to be updated


camclean:
The current code:  
if self.era == "2016":
                self.jesUncertaintyInputFileName = "Summer16_23Sep2016V4_MC_UncertaintySources_" + jetType + ".txt" # Latest is Summer16_07Aug2017_V11 but affected by Formula Evaluator bug

The current recommendation says that we should be using Summer16_07Aug2017_V11, which has updated uncertainties: "Recommended until further notice. The impact of the FormulaEvaluator bug is found to be small compared to the uncertainties." https://twiki.cern.ch/twiki/bin/view/CMS/JECDataMC#Recommended_for_MC

Issue #151, opened 2019-05-28
Jet mass scale corrections in jetmetUncertainties.py are out of date for 2017


camclean:
The current jet mass scale corrections only reflect the values derived for 2016:
self.jmsVals = [1.00, 0.99, 1.01] #nominal, down, up

The 2017 values should be added:
self.jmsVals = [0.982, 0.978, 0.986]

https://twiki.cern.ch/twiki/bin/view/CMS/JetWtagging

Issue #150, opened 2019-05-28
GetSmearValsM() in jetSmearer.py uses wrong smearing values


camclean:
The current nominal, up, and down mass smearing values are old (0.1, 0.2, 0.0):
jet_m_sf_and_uncertainty = dict( zip( [enum_nominal, enum_shift_up, enum_shift_down], [0.1, 0.2, 0.0] ) )

They should be updated to reflect the current measured values (https://twiki.cern.ch/twiki/bin/view/CMS/JetWtagging).

PR #160, opened 2019-06-07
Updated to use latest JME corrections

camclean:
Updated correction values:
- JMS and JMR SFs now reflect W-tagging recommendations. (Addresses issues #150 and #151.)
- Now using latest 2016 MC JECs to address issue #155.

Added raw and jec correction branches to data (to be consistent with MC) and option to redoJEC (or not) to address issue #158.


camclean:
@arizzi @fgolf - are there any problems merging this request?

arizzi:
I let @fgolf comment as he is taking responsibility for nanoaod-tools updates

arizzi:
btw what is the interplay with #163

lcorcodilos:
My fault actually - I just started changing things that I saw were incorrect and it wasn't until I submitted the PR that I started linking my fixes with the issues that @camclean posted. I missed this PR so there's overlap. I'll compare the two and remove what's redundant from mine.

lcorcodilos:
Actually, what's the cleanest/simplest thing to do here? I need to adapt the code from #163 to the code here. I think it makes the most sense to wait for this PR to merge into cms-nanoAOD, merge cms-nanoAOD/master into my fork, and then commit the merge to my fork. I think #163 should update automatically and then only have my relevant changes.

fgolf:
The proposed changes look ok, but I have a question about the scope.  It looks like the updated JMR scale factors come from W-tagging studies, are they also recommended only for a particular use case or generally for other uses and jet types?

camclean:
@fgolf this is a question I plan to discuss at the jetmet meeting tomorrow. It may make more sense to add an option to apply jet mass smearing or not. Either way, the values currently in use are definitely wrong.

lcorcodilos:
@fgolf The discussion today at the JetMET meeting concluded that an option should be added eventually but that a user can factorize what they need with the current setup by taking the raw values and applying each of the saved corrections. 

Once this is merged, I'll submit a new PR with the PUPPI SD mass correction implemented.

PR #162, opened 2019-06-11
Small postprocessor improvements in splitting and logging

gpetruc:
* Better handling of cases with eventRange and maxEntries
* Better reporting of efficiencies and rates


gpetruc:
Also updated with a bugfix from @sscruz to make sure modules are run in the order they are specified in the command line (important if they have dependencies; see https://github.com/gpetruc/nanoAOD-tools/pull/1)

Issue #161, opened 2019-06-07
jet_m_resolution variable undefined in jetSmearer.py


camclean:

eshwen:
Also having this issue. Is someone able to resolve it?

benkrikler:
It seems this has been there since PR #48 when the new functionality was first added by @rappoccio.

camclean:
Should be solved by PR #166

fgolf:
This issue is fixed by PR #167

PR #166, opened 2019-06-19
Use common random number for random smearing of jet pt and mass

IzaakWN:
To be consistent between the central and shifted values, just one random variable should be used. I took advantage of the fact that
```
sigma = jet_pt_resolution*math.sqrt(jet_pt_sf_and_uncertainty[central_or_shift]**2 - 1.)
smearFactor = self.rnd.Gaus(1., sigma)
```
is equivalent to
```
rand = self.rnd.Gaus(0,jet_pt_resolution)
smearFactor = 1. + rand * math.sqrt(jet_pt_sf_and_uncertainty[central_or_shift]**2 - 1.)
```
but now, `rand` does not depend on `jet_pt_sf_and_uncertainty[central_or_shift]` and can be used commonly. This corresponds to the equation on [this TWiki page](https://twiki.cern.ch/twiki/bin/view/CMS/JetResolution#Smearing_procedures). In this PR, it is implemented in both `getSmearValsPt` and a `getSmearValsM`.

Second fix is that `jet_m_resolution` in `getSmearValsM` was previously not defined, causing an error. I am no expert in jet mass smearing, but to fix the error, I simply defined it the same as `jet_pt_resolution` in `getSmearValsPt`. Experts might want to chime in on this.
(This error did not happen before because `jet_m_sf_and_uncertainty[central_or_shift]` was always less than `1.` until commit https://github.com/cms-nanoAOD/nanoAOD-tools/commit/22b99cc24846ce185c534e025131633660448e94.)


AndreasAlbert:
I also observed this bug. If I am not mistaken, you forgot to change the smear function signature [here](https://github.com/cms-nanoAOD/nanoAOD-tools/blob/e56869e78d2eab072e04ce605d6544f90ed12c76/python/postprocessing/modules/jme/jetmetUncertainties.py#L341)

ktht:
I had also implemented this feature in my fork, but my implementation was inspired by the [SmearedJetProducer](https://github.com/cms-sw/cmssw/blob/1fef057ea0118be780f3ccd15eb7b1ce7c9a9dab/PhysicsTools/PatUtils/plugins/SmearedJetProducer.cc) plugin in PatUtils in which the seed is computed from the event level information (run, lumi and event numbers) and from the jet information (specifically, jet eta) as shown [here](https://github.com/cms-sw/cmssw/blob/1fef057ea0118be780f3ccd15eb7b1ce7c9a9dab/PhysicsTools/PatUtils/interface/SmearedJetProducerT.h#L208-L215). It has the advantage that the JER corrections won't depend on the history/order of event processing.

IzaakWN:
@AndreasAlbert, thanks for catching that! It would have introduced a new error for those using groomed jets! Not sure if this PR will get through in its current form, but I fixed it anyway: https://github.com/cms-nanoAOD/nanoAOD-tools/compare/master...IzaakWN:patch-1

@ktht, that sounds like a good idea. Atm the seed is just `12345` for each file. Are you planning to do a PR, or should I add it in this PR or a next one?

ktht:
@IzaakWN I don't think it's a good idea to improve the same feature in two different PRs as someone has to eventually rebase their PR if the other PR gets merged first. So please go ahead and add it here :)

IzaakWN:
Okay, in any case, this PR conflicts with a few other ones in the pipeline. I'll come back to this, if the others go through.

peruzzim:
@fgolf could you please follow up on this PR?

IzaakWN:
Hi @peruzzim & @fgolf, this PR conflicts with PR https://github.com/cms-nanoAOD/nanoAOD-tools/pull/167, which should probably be merged first. I will probably rebase and make a new PR, also implementing a deterministic seed setting, as suggested by @ktht, if those changes are fine with you.

fgolf:
Hi @IzaakWN, yes I was planning to merge PR #167 first if it's ok.  Thanks for waiting and sorry for the long delay.

eshwen:
Hi. Is it possible for someone to follow up with this? We're having this issue https://github.com/cms-nanoAOD/nanoAOD-tools/issues/161 and need the branches from these constructors https://github.com/cms-nanoAOD/nanoAOD-tools/blob/master/python/postprocessing/modules/jme/jetmetUncertainties.py#L485-L492

fgolf:
@IzaakWN PR #167 has been merged, please proceed with your changes.

IzaakWN:
I'll open a new PR.

IzaakWN:
Apologies for noise; reopened this PR, as I managed to resolve the conflicts with merged PR https://github.com/cms-nanoAOD/nanoAOD-tools/pull/167  manually.

This PR now sets a deterministic seed, which is set uniquely for each event in `jetmetUncertaintiesProducer.analyze` with

https://github.com/IzaakWN/nanoAOD-tools/blob/50e03c9accba630af46a5f673cfdb0a20e30e374/python/postprocessing/modules/jme/jetSmearer.py#L60-L68

in analogue to

https://github.com/cms-sw/cmssw/blob/master/PhysicsTools/PatUtils/interface/SmearedJetProducerT.h

@ktht & @camclean, can you check if this implementation is fine with you, please?

python/postprocessing/modules/jme/jetSmearer.py
```diff
@@ -25,52 +25,63 @@ def __init__(self, globalTag, jetType = "AK4PFchs", jerInputFileName = "Spring16
...
         print("Loading JER scale factors and uncertainties from file '%s'" % os.path.join(self.jerInputFilePath, self.jerUncertaintyInputFileName))
         self.jerSF_and_Uncertainty = ROOT.PyJetResolutionScaleFactorWrapper(os.path.join(self.jerInputFilePath, self.jerUncertaintyInputFileName))
-
+        
     def endJob(self):
         pass
-
         
+    
+    def setSeed(self,event):
```
ktht: In [`PhysicsTools/PatUtils/interface/SmearedJetProducerT.h`](https://github.com/cms-sw/cmssw/blob/1fef057ea0118be780f3ccd15eb7b1ce7c9a9dab/PhysicsTools/PatUtils/interface/SmearedJetProducerT.h#L208-L215) that you referred to in the comments, the run number is shifted by 20 bits, lumi number by 10 bits and event number is not shifted at all. I think this is because typically the run number is the smallest, followed by lumi and event number. Shifting them as it's done in C++ code protects the resulting seed from overflowing. Even though the limit on the integer type in Python is ridiculously high, I would still prefer that you implement the same logic of shifting to make the C++ and the Python code equivalent.

One thing I implemented differently wrt the C++ version of the code is that I obtained a different seed not only for every event, but also for every single jet. This is because picking the first jet (by corrected pT) in the event is subject to the effects of JEC, as the order of jets depends on the choice of GT in NanoAOD production.

IzaakWN: Thank, was not paying attention; I commited a change, such that only the run number and lumi block are shifted, not the event number.

I was also think that the order of jets might be an issue. Now two independent users should be able to compare there jets one-to-one, as long as they start from the same JECs (and therefore the same jet order). Do you think it's still better to do it differently than it is now? Where can I see your implementation, is it possible it's private on GitHub?

ktht: Thanks for prompt reply and quick fix in the code.

I have no strong opinion about the seed computation for every single jet, as the effect is negligible. On the other hand, it's not incorrect, computationally expensive or difficult to implement either, so it's up to you if you want to keep the code as it is now, or if you follow my proposal. Here is my implementation for reference:

- run-lumi-event numbers as passed to `getSmearValsPt` and to `getSmearValsPt` functions ([here](https://github.com/HEP-KBFI/nanoAOD-tools/blob/185d56ed23b99a4d86122ecf6f92038d40afa95e/python/postprocessing/modules/jme/jetSmearer.py#L74) and [here](https://github.com/HEP-KBFI/nanoAOD-tools/blob/185d56ed23b99a4d86122ecf6f92038d40afa95e/python/postprocessing/modules/jme/jetSmearer.py#L153)) where the seeds are computed
- the function that computes the seed is [here](https://github.com/HEP-KBFI/nanoAOD-tools/blob/185d56ed23b99a4d86122ecf6f92038d40afa95e/python/postprocessing/modules/jme/jetSmearer.py#L10-L20)

I guess it can be optimized by passing the run-lumi-event numbers first and the call `getSmearVals*` without the run-lumi-event argument. (Once this PR gets merged I'll drop whatever I have implemented in my fork in favor of the implementation in upstream.)

IzaakWN: Using each jet's eta, like in your implementation makes sense, but if you have no strong opinion to include this, I propose to just leave it as it is now. :-p

Issue #164, opened 2019-06-18
Smear with corrected or uncorrected pt?


IzaakWN:
In [this line](https://github.com/cms-nanoAOD/nanoAOD-tools/blob/master/python/postprocessing/modules/jme/jetmetUncertainties.py#L284),
```
... = self.jetSmearer.getSmearValsPt(jet, genJet, rho)
```
the pt of the `jet` object has not been corrected to `jet_pt` (in case it is recalibrated). Should the smearing be done based on the recalibrated jet pt? In that case, one should add
```
jet.pt = jet_pt
... = self.jetSmearer.getSmearValsPt(jet, genJet, rho)
```

IzaakWN:
@camclean, do you know?

camclean:
Hi @IzaakWN, you're right. Thanks for catching that! And same for the mass.

camclean:
I just made a PR. (#170)

PR #170, opened 2019-07-04
Added two missing lines needed to use corrected pt and mass in jet pt and mass smearers

camclean:
Addressed issue #164. Two lines were missing that were needed to access the newly corrected pT and mass when accessing the jet object with the mass and pT smearers.


IzaakWN:
I realized that one other thing might be wrong, which is easily fixed: To propagate the JEC+JER to the MET, `jet_pt_nom - jet_pt` is added to the x and y components. In general, however, `jet_pt` may become a different value from the original jet value, `jet.pt` if the JEC versions before and after nanoAOD processing are different.

If you agree, could you add a few lines, something like
```
jet_pt = jet.pt()
jet_pt_old = jet.pt()
```
to https://github.com/cms-nanoAOD/nanoAOD-tools/blob/bc2f8b5fc3b7b9385981fd05d8dacc5153911e97/python/postprocessing/modules/jme/jetmetUncertainties.py#L277-L279
and change
https://github.com/cms-nanoAOD/nanoAOD-tools/blob/bc2f8b5fc3b7b9385981fd05d8dacc5153911e97/python/postprocessing/modules/jme/jetmetUncertainties.py#L384-L385
to
```
                met_px_nom = met_px_nom - (jet_pt_nom - jet_pt_old)*jet_cosPhi
                met_py_nom = met_py_nom - (jet_pt_nom - jet_pt_old)*jet_sinPhi
```
for example?

camclean:
Daniel Spitzbart is working on an updated way to correct the MET, since this simple subtraction method is not correct.

IzaakWN:
Good to know, in that case that fix would not matter. Thanks for the follow-up!

PR #185, opened 2019-08-20
Adding JMS/JMR corrections to softdrop mass  for tau21DDT<=0.43 + bugfix in JetReCalibrator.py

adewit:
PR with changes in jetmetUncertainties.py made on behalf of @GLP90 who implemented this on a private branch some time ago - PR to avoid diverging from this central repository.

Changes in JetReCalibator.py needed because currently the jme uncertainties module will break if redoJEC=True.

### DIFF
--- a/python/postprocessing/modules/jme/jetmetUncertainties.py
+++ b/python/postprocessing/modules/jme/jetmetUncertainties.py
@@ -142,6 +142,7 @@ class jetmetUncertaintiesProducer(Module):
         if self.doGroomed:
             self.out.branch("%s_msoftdrop_raw" % self.jetBranchName, "F", lenVar=self.lenVar)
             self.out.branch("%s_msoftdrop_nom" % self.jetBranchName, "F", lenVar=self.lenVar)
+            self.out.branch("%s_msoftdrop_tau21DDT_nom" % self.jetBranchName, "F", lenVar=self.lenVar)
             self.out.branch("%s_msoftdrop_corr_JMR" % self.jetBranchName, "F", lenVar=self.lenVar)
             self.out.branch("%s_msoftdrop_corr_JMS" % self.jetBranchName, "F", lenVar=self.lenVar)
             self.out.branch("%s_msoftdrop_corr_PUPPI" % self.jetBranchName, "F", lenVar=self.lenVar)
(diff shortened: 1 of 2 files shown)


python/postprocessing/modules/jme/jetmetUncertainties.py
```diff
@@ -363,6 +375,37 @@ def analyze(self, event):
                 jets_msdcorr_jmrDown.append(jet_pt_jerNomVal *jet_msdcorr_jmrDownVal*jmsNomVal  *jet_msdcorr_raw)
                 jets_msdcorr_jmsUp  .append(jet_pt_jerNomVal *jet_msdcorr_jmrNomVal *jmsUpVal   *jet_msdcorr_raw)
                 jets_msdcorr_jmsDown.append(jet_pt_jerNomVal *jet_msdcorr_jmrNomVal *jmsDownVal *jet_msdcorr_raw)
+
+                #Also evaluated JMS&JMR SD corr in tau21DDT region: https://twiki.cern.ch/twiki/bin/viewauth/CMS/JetWtagging#tau21DDT_0_43
+                tmp_jmsVals = self.jmsVals
```
fgolf: This is ok, but it looks like it isn't necessary to assign and restore to jmsVals/jmrVals.  What is written only makes use of the values locally here.

python/postprocessing/modules/jme/jetmetUncertainties.py
```diff
@@ -363,6 +375,37 @@ def analyze(self, event):
...
+                    self.jmsVals = [1.014,1.007,1.021]
+                    self.jmrVals = [1.086,1.176,0.996]
+                elif self.era in ["2017"]:
+                    self.jmsVals = [0.983,0.976,0.99]
+                    self.jmrVals = [1.080,1.161,0.999]
+                jmsNomVal = self.jmsVals[0]
+                jmsDownVal = self.jmsVals[1]
+                jmsUpVal = self.jmsVals[2]
+
+                ( jet_msdcorr_tau21DDT_jmrNomVal, jet_msdcorr_tau21DDT_jmrUpVal, jet_msdcorr_tau21DDT_jmrDownVal ) = self.jetSmearer.getSmearValsM(groomedP4, genGroomedJet) if groomedP4 != None and genGroomedJet != None else (0.,0.,0.)
```
fgolf: IIUC, jetSmearer.getSmearValsM is called with a different set of jmrVals than the ones you've defined a few lines above.  Is this intended or should it be called using the new jmrVals?

adewit: Thanks for spotting this oversight, it should have been called with the new set of jmrVals. I've now changed it such that the jetSmearer.jmr_vals gets set with the required values before the evaluation. Another option would be to initialise a separate jetSmearer for these different jmrValues, if desired I can change it to that. Just let me know.

PR #187, opened 2019-08-24
remove directory tree listing

belforte:
Most users use this script as-is and end un with very long output where the full directory three is printed in each job. Long outputs put strain on submission infrastructure and CRAB servers. Users should only do this when really needed, and for a few jobs. Listing same directory in each job in a task is totally useless.

### DIFF
--- a/crab/crab_script.py
+++ b/crab/crab_script.py
@@ -10,5 +10,4 @@ p=PostProcessor(".",inputFiles(),"Jet_pt>200",modules=[exampleModuleConstr()],pr
 p.run()
 
 print "DONE"
-os.system("ls -lR")
 


swertz:
FYI I think this should be removed too: https://github.com/cms-nanoAOD/nanoAOD-tools/blob/master/crab/crab_script.sh#L8

belforte:
> FYI I think this should be removed too: https://github.com/cms-nanoAOD/nanoAOD-tools/blob/master/crab/crab_script.sh#L8
ok with me: https://github.com/cms-nanoAOD/nanoAOD-tools/pull/190

belforte:
actually most of https://github.com/cms-nanoAOD/nanoAOD-tools/blob/master/crab/crab_script.sh
is useless since CRAB wrapper already expands the user sandbox in $CMSSW_BASE
since CRAB wrapper creates a CMSSW scram area (cmsrel + cmsenv + untar sandbox) also when using scriptExe.

PR #193, opened 2019-10-02
Updates to run V19 2018 JECs and corresponding JERs

kschweiger:
This PR adds the new JEC and JER files (see. [Hypernews](https://hypernews.cern.ch/HyperNews/CMS/get/jes/898.html)) as well as other changes  to run them (and bug fixes).

# Changes:
1. JEC and JER files were provided as uncompressed tar's. Since the post processor code expects them to be compressed files (so. `tgz` instead of `.tar`) I repackaged the files. Procedure is detailed in `data/jme/README.md`.
    - The data JEC file did not exhibit the same structure as previous versions. Details on repackaging are given in `data/jme/README.md`.
   - If there are strong objections to this I can change to uncompressed files but this requires further changes down the line because the extraction part is hardcoded (as well as file extension)
2. Updated `jetRecalib.py`, `jetSmearer.py`, `jetmetUncertainties.py` and `jetmetHelperRun2.py` to use the new files
3. Added options to disable non-standard behavior in jetmet modules:
    - `jetmetUncertainties.py`: Parameter `doWTagMassRes` (set to `False` by default) is used to set how the jet mass will be corrected. If `False`, the pt factors will be used for mass. The previous default behavior is restored when passing `True`.
    - `jetRecalib `: Parameter `doPUPPIMassCorr ` (set to `False` by default) is used to activate PUPPI SD mass correction
4. Minor bugfixes:
   - `jetRecalib` : Failed when running with AK4 jets, because  `self.subJetBranchName` was not set in that case 
   - `jetmetHelperRun2.py`: Variable use for `archive` in `jetRecalib` call was not defined
 5. Removed JEC and JER files that are not used anymore 

# Additional note
Did **not update** `jmrValues` and `jmsValues` in `jetmetHelperRun2.py` since I'm no expert on this topic (but added a warning for this 😉)


kschweiger:
Addressed with PR #191 . Closing this one.

PR #194, opened 2019-10-07
Backfill Run tree in haddnano to support multi-point signal scans

peruzzim:
Backfilling also the Run tree in both boolean and non-boolean branches.
This is needed to support multi-point signal scans (cms-nanoAOD/cmssw#398).

To be tested thoroughly


peruzzim:
This is needed to merge NanoAODv6, I merge it, please report in case you observe issues at merging.

PR #197, opened 2019-10-16
Fix bug in large pileup weight handling

AndrewLevin:
Currently, the cropped vector will grow larger and larger in each iteration, while it should always have a fixed length or be empty.


fgolf:
Accepting this PR to fix the nan problem and we'll follow-up on proposed changes in PR #87 .

PR #202, opened 2019-10-27
FatJetUncertainties: Do not create unused branches

AndreasAlbert:
The branches:

`FatJet_corr_JER`
`FatJet_corr_JMS`
`FatJet_corr_JMR`

are only ever filled for MC, but are currently also set up for data. This PR makes it so that they are only created for MC.

@danbarto what do you think?


fgolf:
discussed with @AndreasAlbert and agree to keep these branches and set the defaults to values corresponding to no effect

Issue #219, opened 2019-12-18
Output Branch Selection disables input branches as well 


lenzip:
I have noticed that when running a chain of modules in the postprocessor with an outputbranchselection only the first module that is run has access to the full event, the second module has access only to the branches that are in the outputbranchselection. I believe this is not the intended behaviour, and that the outputbranchselection should be applied only upon writing the output file. I believe the issue is in those two lines  https://github.com/cms-nanoAOD/nanoAOD-tools/blob/master/python/postprocessing/framework/output.py#L86-L87, which should not be there (the only place where outputbranchselection should play a role is here: https://github.com/cms-nanoAOD/nanoAOD-tools/blob/master/python/postprocessing/framework/output.py#L129).
The friend tree output is also affected by the same behaviour.

fgolf:
Fixed by #222

PR #222, opened 2020-01-14
Fix to output branch selection

lenzip:
Addresses issue #219


fgolf:
I agree with change you made, but I haven't been able to reproduce the problematic behavior you report in #219.  I'll merge this, but could you send me an example that reproduces the problem?  Thanks.

PR #223, opened 2020-01-18
Added JMS and JMR sf for FatJet for 2018 in jetmetHelper

valsdav:
Updated 2018 JMS and JMR scale factors for FatJets from https://twiki.cern.ch/twiki/bin/view/CMS/JetWtagging

### DIFF
--- a/python/postprocessing/modules/jme/fatJetUncertainties.py
+++ b/python/postprocessing/modules/jme/fatJetUncertainties.py
@@ -383,11 +383,16 @@ class fatJetUncertaintiesProducer(Module):
                         jmstau21DDTDownVal = 1.007
                         jmstau21DDTUpVal = 1.021
                         self.jetSmearer.jmr_vals = [1.086,1.176,0.996]
-                    elif self.era in ["2017","2018"]:
+                    elif self.era in ["2017"]:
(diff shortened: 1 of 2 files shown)


fgolf:
The proposed changes are ok, but before merging them I'd like to understand why the numbers in fatJetUncertainties.py and jetmetHelperRun2.py.  Should they be the same?

valsdav:
Hi, thanks for your comment. 
As far as I have understood the numbers in jetmetHelperRun2 correspond to tau21<0.45 HP category official WP, whereas the ones inside the fatJetUncertainty module correspond to the tau21DDT working point. 

I agree that this could be refactored to save both in the jetmethelper for clarity.

fgolf:
@camclean Can you take a look at comment on this?  Thanks.

camclean:
"As far as I have understood the numbers in jetmetHelperRun2 correspond to tau21<0.45 HP category official WP, whereas the ones inside the fatJetUncertainty module correspond to the tau21DDT working point."
>> This is correct. The numbers look right.

It would indeed be nice to refactor the modules so that the values were all in the jetmethelper.

fgolf:
Thanks @camclean.  I'll merge this and open an issue to factorize the remaining values into jetmethelper.

PR #232, opened 2020-02-06
Additions to jet met tools

swertz:
Feedback wanted: we might not want all of this in the central tools - however for current analyses I think these additions would be useful. Sorry for mixing several changes in the same PR:

- Add several options to `createJMECorrector()` (and corresponding changes in `jetmetUncertaintiesProducer`):
   - HEM issue: Following https://hypernews.cern.ch/HyperNews/CMS/get/JetMET/2000.html, this adds a jes uncertainty called `HEMissue` where the "down" variation scales the jet pt down according to the values suggested in the HN (the "up" variation is equal to the nominal).
   - Split JER uncertainties: Following https://twiki.cern.ch/twiki/bin/view/CMS/JetResolution#Run2_JER_uncertainty_correlation, this splits the jer uncertainties into 6 components (called `jer0...jer5`) depending on jet pt/eta, and propagates this to the MET. If the flag is left `False`, the current behaviour is unchanged. Here I'm actually not sure if I propagate correctly to the `METFixEE2017`: feedback welcome.
   - "Regrouped" JEC uncertainties, following https://twiki.cern.ch/twiki/bin/view/CMS/JECUncertaintySources#Run_2_reduced_set_of_uncertainty. Specifying `jesUncert="Merged"` (instead of `"All"`) to `createJMECorrector()` will use the reduced set of uncertainties instead of the whole set. The corresponding txt files from the twiki were added to the archives for all 3 years.
- Update 2018 JER to V7b (probably useless here?).
- Fix "bug" in jet smearer: the clipping of small values (in case the smeared jet energy is negative, for stochastic smearing) was not 100% consistent with the CMSSW implementation: https://github.com/cms-sw/cmssw/blob/master/PhysicsTools/PatUtils/interface/SmearedJetProducerT.h#L280


python/postprocessing/modules/jme/jetmetUncertainties.py
```diff
@@ -380,33 +428,52 @@ def resolution_matching(jet, genjet):
...
+                  if jesUncertainty == "HEMIssue":
+                      delta = 1.
+                      if jet.phi > -1.57 and jet.phi < -0.87:
+                          if jet.eta > -2.5 and jet.eta < -1.3:
+                            delta = 0.8
+                          elif jet.eta <= -2.5 and jet.eta > -3:
+                            delta = 0.65
+                      jet_pt_jesUp[jesUncertainty] = jet_pt_nom
+                      jet_pt_jesDown[jesUncertainty] = delta * jet_pt_nom
+                      jet_mass_jesUp[jesUncertainty] = jet_mass_nom
```
AndreasAlbert: Should the delta also be applied to the jet mass? This is done for "regular" JES in the code below

AndreasAlbert: Ah, never mind, jes up == nominal here. I see it is applied for the jes down.

swertz: Right, for the HEM issue the variation is single-sided.

python/postprocessing/modules/jme/jetmetUncertainties.py
```diff
@@ -447,10 +515,14 @@ def resolution_matching(jet, genjet):
...
-              met_py_jerUp += delta_y_rawJet - met_unclEE_y
-              met_px_jerDown += delta_x_rawJet - met_unclEE_x
-              met_py_jerDown += delta_y_rawJet - met_unclEE_y
+              
+              # FIXME is this correct?
+              for jerID in self.splitJERIDs:
+                  met_px_jerUp[jerID] += delta_x_rawJet - met_unclEE_x
+                  met_py_jerUp[jerID] += delta_y_rawJet - met_unclEE_y
+                  met_px_jerDown[jerID] += delta_x_rawJet - met_unclEE_x
+                  met_py_jerDown[jerID] += delta_y_rawJet - met_unclEE_y
```
AndreasAlbert: This loop seems to be consistent with what the code did here before your changes, right? So the question is not so much whether your change is consistent, but whether the method was already consistent before?

AndreasAlbert:
On this PR in general: I think the changes made here are excellent to have in the central repository. We should definitely include this to avoid duplication of work.

Regarding validation: It seems to me that unless one uses special arguments, the behavior of the uncertainty calculation is unchanged with the small exception of the negative value clipping. Is this correct? If it is, I would advocate to merge this PR quickly, once the EE question is settled. The extra features like HEM and JER splits will have to be looked at carefully by the people who want to use them for physics studies, but that can happen afterwards.

python/postprocessing/modules/jme/jetmetUncertainties.py
AndreasAlbert: paging also @danbarto

swertz: Right, I had misunderstood the recipe somehow, it should be fine as it is... (I'll remove the FIXME from the file)

swertz:
> Regarding validation: It seems to me that unless one uses special arguments, the behavior of the uncertainty calculation is unchanged with the small exception of the negative value clipping. Is this correct? 

Yes, that was the idea. I've checked that it is the case (on a single file and for some of the variations).

python/postprocessing/modules/jme/jetmetUncertainties.py
danbarto: The v2 recipe part should in fact only run when you start from raw MET (so I guess what's now under redoJEC=True) - did you check that this is done correctly?

swertz: Hm, looking over it again I think the `redoJEC` flag is pretty much useless here and could be removed entirely.

About the v2 recipe, I'm not 100% sure what's being done and whether it is correct, e.g. it depends on what's given as `globalTagProd` option, and that is never set when calling from `jetmetHelperRun2.py`...

But this PR should be consistent with what is currently there in any case.

python/postprocessing/modules/jme/jetmetUncertainties.py
```diff
@@ -113,8 +142,9 @@ def beginJob(self):
             jesUncertainty_label = jesUncertainty
             if jesUncertainty == 'Total' and len(self.jesUncertainties) == 1:
```
AndreasAlbert: Does this line still work with the changes in this PR? If I'm running with hemfix enabled and want total JES unc, the JetCorrectorParameters constructor a few lines below segfaults. I believe this is because in this case `len(self.jesUncertainties)` is `2`, which means that `jesUncertainty_label` will not be modified here. If `jesUncertainty` is `'Total'`, then `jesUncertainty_label` will also be `'Total'`, while it should be `''`. 

Long story short, it seems to me the that the line should be:

```python
if jesUncertainty == "Total" and (len(self.jesUncertainties)==1 or (len(self.jesUncertainties)==2 and "HEMIssue" in self.jesUncertainties)):
                jesUncertainty_label = ""
```

python/postprocessing/modules/jme/jetmetUncertainties.py
```diff
@@ -366,8 +407,15 @@ def resolution_matching(jet, genjet):
                 jets_mass_nom    .append(jet_mass_nom)
 
             if not self.isData:
-              jet_pt_jerUp         = jet_pt_jerUpVal  *jet_pt
-              jet_pt_jerDown       = jet_pt_jerDownVal*jet_pt
+              jet_pt_jerUp = { jerID: jet_pt_nom for jerID in self.splitJERIDs }
+              jet_pt_jerDown = { jerID: jet_pt_nom for jerID in self.splitJERIDs }
+              jet_mass_jerUp = { jerID: jet_mass_nom for jerID in self.splitJERIDs }
```
AndreasAlbert: This line crashes for me because `jet_mass_nom` is only defined for `iJet < nJet`, i.e. only for high-pt jets, but the overall loop also runs over low-pt jets. Suggest to either replace by `jet_pt_jerNomVal*jet_mass if self.applySmearing else jet_mass` from L404, or simply move the `jet_mass_nom` definition out of the if block.

AndreasAlbert: I think this is actually tricky to reproduce because it will only happen if the first event has `nJet==0`. In other cases, there is no crash, you just get reuse the `jet_mass_nom` value from the previous iteration of the loop, which is worse than crashing

swertz: Good catch, should be fixed now.

python/postprocessing/modules/jme/jetmetUncertainties.py
swertz: Indeed, I hadn't checked that particular combination... should be OK now.

swertz:
I've checked again the results between the previous code and with this PR, and while the jets always agree 100%, there are a few events where the MET ends up different (whether the smearing is applied or not)... I'm quite puzzled as to where this would come from.

python/postprocessing/modules/jme/jetmetUncertainties.py
```diff
@@ -260,18 +298,18 @@ def resolution_matching(jet, genjet):
...
             else:
                 jet_rawpt = -1.0 * jet_pt #If factor not present factor will be saved as -1
                 jet_rawmass = -1.0 * jet_mass #If factor not present factor will be saved as -1
 
-            (jet_pt, jet_mass)    = self.jetReCalibrator.correct(jet,rho)
-            (jet_pt_l1, jet_mass_l1) = self.jetReCalibratorL1.correct(jet,rho)
+            (jet_pt, jet_mass) = self.jetReCalibrator.correct(jet,rho)
             jet.pt = jet_pt
             jet.mass = jet_mass
+
```
AndreasAlbert: What is the reasoning behind the change in order here? Previously, `jetRecalibrator1` was called *before* `jet.pt` was set, now it's called *after* `jet.pt` is set. Doesn't that make a difference (at least when you're switching JES)?

AndreasAlbert:
> I've checked again the results between the previous code and with this PR, and while the jets always agree 100%, there are a few events where the MET ends up different (whether the smearing is applied or not)... I'm quite puzzled as to where this would come from.

So you're saying that `MET_pt_nom` disagrees, while all `Jet_pt_nom` agree? Have you checked the low pt jets? A difference could be hiding there.

python/postprocessing/modules/jme/jetmetUncertainties.py
swertz: Unintented (had made another change and only partly reverted it)... But now I understand the issue with the MET: while this had no effect on regular jets if the same JES is used, it did mess up the low-pt jets and hence the corrected MET.

Now I get perfect agreement with the previous version of the code also for the MET.

swertz:
> So you're saying that `MET_pt_nom` disagrees, while all `Jet_pt_nom` agree? Have you checked the low pt jets? A difference could be hiding there.

Indeed that was the case, it's fixed now.

AndreasAlbert:
Great! Thanks for following up!

AndreasAlbert:
Regarding the removal of `redoJEC`: Why do we not need it? Isn't there still a use case for people wanting JES unc without re-calculating JEC?

swertz:
> Regarding the removal of `redoJEC`: Why do we not need it? Isn't there still a use case for people wanting JES unc without re-calculating JEC?

It wasn't used in the code, so removing it is at least consistent with what's being done.

We could re-add it for the case you mention, but that would require a few more changes, in particular:
- Currently MET always starts from raw MET and type-1 corrections are recalculated.
- Low-pt jets always start from the raw pt (for some reason...?) so they always need the JEC to be applied.
Since re-calculating the JEC is a small overhead compared to all the uncertainties, perhaps we could leave things as they are to keep the code readable...?

AndrewLevin:
Hi, I am interested in getting these features for puppimet. I previously made a pull request to add the puppimet JER and JES uncertainties directly to the NANOAOD: https://github.com/cms-nanoAOD/cmssw/pull/481. I am wondering if I should add more variables like that such as `PuppiMET_HEMissueUp`. @ahinzmann @peruzzim

ahinzmann:
adding @camclean as well

ahinzmann:
> Hi, I am interested in getting these features for puppimet. I previously made a pull request to add the puppimet JER and JES uncertainties directly to the NANOAOD: [cms-nanoAOD/cmssw#481](https://github.com/cms-nanoAOD/cmssw/pull/481). I am wondering if I should add more variables like that such as `PuppiMET_HEMissueUp`, `PuppiMET_JER1Up`, and `PuppiMET_JES1Up`. @ahinzmann @peruzzim

I think what we can realistically support for pre-UL NanoAODs is the following:
1. single JEC variation for PUPPI MET in NanoAOD
2. single JER variation for PUPPI MET in NanoAOD
3. compute with nanoAOD-tools split JEC into all sources or reduced set of sources for CHS jets and PF MET
4. compute with nanoAOD-tools split JER into pt/eta sources for CHS jets and PF MET
(3. and 4. addressed by this PR)

Analyses that are sensitive to JEC+JER and need to split JEC and JER sources for PUPPI MET can be done with MiniAOD for now (not mature enough for Nano and requires many new floats) and better be discussed in a MET meeting as this is something we have not studied in JetMET yet. Input is very welcome!!

AndrewLevin:
ok, and what about the `PuppiMET_HEMissueUp`? that may be more important because it is not a component of the total JEC and JER variations which we have already added

ahinzmann:
The HEMissueUp is just a temporary recipe to check impact of the effect, not an actual uncertainty estimate. The proper JEC with HEM calibrated is still to come. For estimating the size of the effect of HEM JECs that check can be performed with CHS jets on top of NanoAOD.

python/postprocessing/modules/jme/jetmetUncertainties.py
```diff
@@ -380,33 +428,52 @@ def resolution_matching(jet, genjet):
...
-                  self.jesUncertainty[jesUncertainty].setJetPt(jet_pt_nom)
-                  self.jesUncertainty[jesUncertainty].setJetEta(jet.eta)
-                  delta = self.jesUncertainty[jesUncertainty].getUncertainty(True)
-                  jet_pt_jesUp[jesUncertainty]   = jet_pt_nom*(1. + delta)
-                  jet_pt_jesDown[jesUncertainty] = jet_pt_nom*(1. - delta)
+
+                  # cf. https://hypernews.cern.ch/HyperNews/CMS/get/JetMET/2000.html
+                  if jesUncertainty == "HEMIssue":
+                      delta = 1.
+                      if jet.phi > -1.57 and jet.phi < -0.87:
```
camclean: from the recipe: "(all jets with pt>15 GeV passing the tight ID -in order to reject muons/electrons-, and propagate this to the MET as well)"
>> Where is the tight ID applied?

python/postprocessing/modules/jme/jetmetHelperRun2.py
```diff
@@ -88,15 +88,15 @@ def createJMECorrector(isMC=True, dataYear=2016, runPeriod="B", jesUncert="Total
...
     # no MET variations calculated
     else:
       if isMC:
-          jmeCorrections = lambda : fatJetUncertaintiesProducer(era=dataYear,                      globalTag=jecTag_, jesUncertainties=jmeUncert_, redoJEC=redojec, jetType = jetType, jerTag=jerTag_, jmrVals = jmrValues_, jmsVals = jmsValues_, applySmearing = applySmearing)
+          jmeCorrections = lambda : fatJetUncertaintiesProducer(era=dataYear,                      globalTag=jecTag_, jesUncertainties=jmeUncert_, jetType = jetType, jerTag=jerTag_, jmrVals = jmrValues_, jmsVals = jmsValues_, applySmearing = applySmearing)
       else:
-          jmeCorrections = lambda : fatJetUncertaintiesProducer(era=dataYear, archive=archiveTag_, globalTag=jecTag_, jesUncertainties=jmeUncert_, redoJEC=redojec, jetType = jetType, jerTag=jerTag_, jmrVals = jmrValues_, jmsVals = jmsValues_, isData=True)
+          jmeCorrections = lambda : fatJetUncertaintiesProducer(era=dataYear, archive=archiveTag_, globalTag=jecTag_, jesUncertainties=jmeUncert_, jetType = jetType, jerTag=jerTag_, jmrVals = jmrValues_, jmsVals = jmsValues_, isData=True)
```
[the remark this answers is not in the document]
swertz: The recipe makes no mention of AK8 jets...

swertz:
Updated, indeed I had neglected the fat jets. It should at the very least not break things now, but I don't know about "merged" uncertainties or the HEM fix for those...

python/postprocessing/modules/jme/jetSmearer.py
```diff
@@ -144,13 +138,14 @@ def getSmearValsPt(self, jetIn, genJetIn, rho):
...
+          for central_or_shift in [ enum_nominal, enum_shift_up, enum_shift_down ]:
             # check that smeared jet energy remains positive,
             # as the direction of the jet would change ("flip") otherwise - and this is not what we want
-            if (smearFactor*jet.Perp()) < 1.e-2:
-                smearFactor = 1.e-2
-            smear_vals[central_or_shift] = smearFactor
-        
+            if smear_vals[central_or_shift]*jet.E() < 1.e-2:
+               smear_vals[central_or_shift] = 1.e-2 / jet.E()
+
```
camclean, line 148: Please decrease the indent of this block so that "for central_or_shift in [ enum_nominal, enum_shift_up, enum_shift_down ]:" is in line with "if genJet:" and the corresponding "else:". Currently you are only applying this clipping to the jets without a matched genJet. However, it should be applied to all jets.

camclean:
@swertz, can you update this PR?

python/postprocessing/modules/jme/jetmetUncertainties.py
```diff
@@ -66,16 +72,20 @@ def __init__(self, era, globalTag, jesUncertainties = [ "Total" ], archive=None,
 
         if len(jesUncertainties) == 1 and jesUncertainties[0] == "Total":
             self.jesUncertaintyInputFileName = globalTag + "_Uncertainty_" + jetType + ".txt"
+        elif jesUncertainties[0] == "Merged" and not self.isData:
```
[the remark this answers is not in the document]
swertz: Should it also be done for `AK4PFPuppi` and `AK8PFchs`?

swertz:
I've added the changes for AK8 jets, sorry for the delay.

While running some checks again I noticed something odd: in 2016 data, the recorrected jet pt (`jet_pt_nom`) is different from the jet pt stored in nanoAOD (`jet_pt`), even though the JEC should be the same:
image

In 2017 and 2018 data there is no change, as expected (apart from small variations, probably due to different implementation+reduced numerical precision?):
image

Recorrected fat jets in 2016 data have a ~2% offset as well. Note that I observe the same thing without the changes introduced in this PR.

camclean:
The values should not be identical since the jets are smeared by the JER corrections applied. However, the shift is indeed odd.

Looking at the changes you've made, I'm happy with everything. @ahinzmann?

ahinzmann:
> The values should not be identical since the jets are smeared by the JER corrections applied. However, the shift is indeed odd.
> 
> Looking at the changes you've made, I'm happy with everything. @ahinzmann?

Well, the PR looks fine, but is the shift in JEC understood? Do you understand from what change it originates?

camclean:
I can check where it comes from, but I don't think that should delay this PR.

ahinzmann:
> I can check where it comes from, but I don't think that should delay this PR.

ok, agree. It would be good to check where it comes from, just to make sure this is related to some issue in Mini or Nano.

swertz:
> The values should not be identical since the jets are smeared by the JER corrections applied. However, the shift is indeed odd.

The plots are data, so there is no JER smearing. The effect might be present in MC too, I haven't checked since I was applying the smearing there.

Anyway, as you said this isn't related to this PR.

camclean:
So @ahinzmann @fgolf are we good to merge this PR?

ahinzmann:
yes

swertz:
So can this be merged?

swertz:
Would it be possible to merge this before #244 ?

swertz:
Yet another ping...

gouskos:
Hi @swertz Can you please move the updated correction files like data/jme/Autumn18_V19_MC.tgz and similar in a new PR [which can be merged faster] and in this PR leave only changes in the code itself which requires more review? thank you.

swertz:
Hi, I could, but the simplest would simply to merge this PR... Hasn't it been reviewed already?

gouskos:
Hi @swertz , I have the following comments/suggestions on this PR:
(a) It would be useful the tool to have the functionality to apply the AK4 JEC, JER on the subjets returned by the softdrop subjets [unless I missed it]. If this would require time to implement can be carried out in a separate PR
(b) I believe it would be useful to add the flexibility to the analyser to store the desired variation [e.g. JEC up/down and not JER etc..
(c) I did a quick check and the outputs look reasonable - For completeness, do you have a set of plots/studies used in JME for the validation of the tool to attach in the PR ?

gouskos:
To summarise discussions over emails:
We will merge this PR since it includes many bug fixes and then there will be a follow-up with JME (sub-)conveners on suggestions, like the ones below: 
(a) It would be useful the tool to have the functionality to apply the AK4 JEC, JER on the subjets returned by the softdrop subjets [unless I missed it]. If this would require time to implement can be carried out in a separate PR
(b) I believe it would be useful to add the flexibility to the analyser to store the desired variation [e.g. JEC up/down and not JER etc..
(c) I think the JMS/JMR option has to be re-structured. Currently these values [well place-holder for now] are supposed to be for a specific WP for one of the taggers [tau21 IIRC]. I think we need to include the flexibility to the analyzer to include other WP and/or taggers.

PR #233, opened 2020-02-06
B-tagging scale factors: multiple working points or taggers; specify list of JEC uncertainties

swertz:
Currently, the `btagSFProducer` module will only add the SFs for a single working point (hardcoded to `M`). Furthermore, the naming of the branches (`Jet_btagSF`) prevents the user to add SFs for several taggers. Finally, the shape correction SFs include systematic variations that have to be correlated with the JEC variations. By default only the total uncertainty is taken into account, but in case one uses SFs which include the split (or reduced) JEC uncertainty sources, one would like to have all the corresponding SF variations stored as well. 

With this PR, the user can select all the working points he'd like (default: `['M', 'shape_corr']`), and the branches are now named as e.g. `Jet_btagSF_<tagger>_<WP>` so that several modules can be added to get the SFs for different taggers.

Also, an extra option is added to specify the list of JEC uncertainties to be considered for b-tagging shape correction SFs.

### DIFF
--- a/python/postprocessing/modules/btv/btagSFProducer.py
+++ b/python/postprocessing/modules/btv/btagSFProducer.py
@@ -1,43 +1,44 @@
 import ROOT
 import os
+from itertools import chain
+
 ROOT.PyConfig.IgnoreCommandLineOptions = True
 
 from PhysicsTools.NanoAODTools.postprocessing.framework.datamodel import Collection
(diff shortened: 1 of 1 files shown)


fgolf:
Thanks @swertz, this is a nice improvement.  I added a couple sentences to [the wiki](https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookNanoAOD#POG_specific) describing the change.

PR #240, opened 2020-05-25
JES uncertainty fix

danbarto:
There is a discrepancy of the propagation of JES variations to type-1 MET between CMSSW and NanoAOD-tools. Discussed here:
https://indico.cern.ch/event/921037/contributions/3869620/attachments/2042217/3420993/nanoaod_jes_talk_v2.pdf
The fix was tested for 2017 (with METv2 recipe) and 2018. Consistent results up to the numerical precision of nanoAOD and the following bin migrations are now achieved.

### DIFF
--- a/python/postprocessing/modules/jme/jetmetUncertainties.py
+++ b/python/postprocessing/modules/jme/jetmetUncertainties.py
@@ -510,8 +510,6 @@ class jetmetUncertaintiesProducer(Module):
           self.out.fillBranch("%s_pt_unclustEnDown" % self.metBranchName, math.sqrt(met_px_unclEnDown**2 + met_py_unclEnDown**2))
           self.out.fillBranch("%s_phi_unclustEnDown" % self.metBranchName, math.atan2(met_py_unclEnDown, met_px_unclEnDown))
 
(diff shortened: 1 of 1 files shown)


danbarto:
Just to clarify: This fix only concerns JES. The closure of JER uncertainties/variations of miniAOD/nanoAOD is still under investigation.

AndreasAlbert:
Thanks @danbarto!

IIUC, the main issue fixed here is the selection of jets for which the uncertainty is propagated to MET, i.e. [this line here](https://github.com/cms-nanoAOD/nanoAOD-tools/pull/240/files#diff-efe14f8220f1ffc7318e6a301464109aR412). The way it is implemented right now, the exact same set of jets is used for 
a) calculating T1 MET and JES uncertainties
b) calculating T1 smeared MET and JER uncertainties

Is that is the right approach in principle? If maybe someone from JME knows the answer for certain, that could guide the further investigation of the JER-related behavior.

danbarto:
To be honest, I'm not entirely sure what should be done if JER smeared jets are used in type-1 MET. The current implementation still resembles the "original" nanoAOD-tools implementation, just with proper type-1 corrections for JES.
Maybe @ahinzmann or @lathomas can comment?

PR #250, opened 2020-09-17
Fix a typo at line 52

fleble:
Hi @gouskos 
Correcting a typo at line 52, missing "," at the end of the line I think


gouskos:
> Hi @gouskos
> Correcting a typo at line 52, missing "," at the end of the line I think

Thanks @fleble

PR #255, opened 2020-10-29
Typo fix (#6)

Laurits7:
As @andreypz pointed out in the comment of PR cms-nanoAOD#251, then there was a mistake in one of the print statements that was introduced in that PR. Added the str() to the print statement and made some further pep8.

### DIFF
--- a/python/postprocessing/framework/postprocessor.py
+++ b/python/postprocessing/framework/postprocessor.py
@@ -42,9 +42,9 @@ class PostProcessor:
                 "hadd. No name specified for the output file, will use tree.root")
             self.haddFileName = "tree.root"
         self.branchsel = BranchSelection(branchsel) if branchsel else None
-        if outputbranchsel != None:
+        if outputbranchsel is not None:
             self.outputbranchsel = BranchSelection(outputbranchsel)
-        elif outputbranchsel == None and branchsel != None:
+        elif outputbranchsel is None and branchsel is not None:
             # Use the same branches in the output as in input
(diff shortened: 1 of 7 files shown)


mariadalfonso:
@andreypz @camclean 
Please check if this PR fixes the two issues you reported.
notify the outcome here.

camclean:
Looks good. Thanks!

andreypz:
I did not have time to check this, but looking at the code, it should be fixed.
In general, someone should just check that [crab submission](https://github.com/cms-nanoAOD/nanoAOD-tools/tree/master/crab) still works after this PR (in CMSSW using python2).

python/postprocessing/framework/postprocessor.py
```diff
@@ -127,10 +127,10 @@ def run(self):
                     "Running with --noout and no modules does nothing!")
 
         # Open histogram file, if desired
-        if (self.histFileName != None and self.histDirName == None) or (self.histFileName == None and self.histDirName != None):
+        if (self.histFileName is not None and self.histDirName is None) or (self.histFileName is None and self.histDirName is not None):
             raise RuntimeError(
                 "Must specify both histogram file and histogram directory!")
-        elif self.histFileName != None and self.histDirName != None:
+        elif self.histFileName is not None and self.histDirName is None:
```
nurfikri89, line 133: Hi @Laurits7,  shouldn't this be  
```python
self.histDirName is not None
```
at the end?

nurfikri89: Fixed in PR #263

PR #244, opened 2020-07-08
Updates on T1Smear MET calculation: Central value + JME uncertainties

alpakpinar:
This PR contains new code that has:

- Separate calculations of JME uncertainties for T1 and T1Smear MET, which are stored in different branches
- An option flag in jetmetHelper to specify which MET uncertainties (T1 or T1Smear) are to be saved

PR #256, opened 2020-11-02
Fix JES uncertainties on T1(smear) MET with split JER (indentation)

pieterdavid:
Since the loop over JES uncertainty sources was put inside the one over JER bins in https://github.com/cms-nanoAOD/nanoAOD-tools/pull/244, the JES variations are applied six times instead of once if JER uncertainty splitting is enabled.
I just removed the indentation, but I guess this also allows for some of the line splitting to be reverted - let me know if there's some script/guideline for this.

### DIFF
--- a/python/postprocessing/modules/jme/jetmetUncertainties.py
+++ b/python/postprocessing/modules/jme/jetmetUncertainties.py
@@ -729,28 +729,28 @@ class jetmetUncertaintiesProducer(Module):
                                     jerID] - (jet_pt_L1L2L3 * jerDownVal -
                                               jet_pt_L1) * jet_sinPhi
 
-                                # Calculate JES uncertainties on unsmeared MET
-                                for jesUncertainty in self.jesUncertainties:
-                                    met_T1_px_jesUp[
-                                        jesUncertainty] = met_T1_px_jesUp[
-                                            jesUncertainty] - (
-                                                jet_pt_jesUpT1[jesUncertainty]
-                                                - jet_pt_L1) * jet_cosPhi
-                                    met_T1_py_jesUp[
-                                        jesUncertainty] = met_T1_py_jesUp[
-                                            jesUncertainty] - (
-                                                jet_pt_jesUpT1[jesUncertainty]
-                                                - jet_pt_L1) * jet_sinPhi
-                                    met_T1_px_jesDown[
-                                        jesUncertainty] = met_T1_px_jesDown[
-                                            jesUncertainty] - (
-                                                jet_pt_jesDownT1[jesUncertainty]
-                                                - jet_pt_L1) * jet_cosPhi
-                                    met_T1_py_jesDown[
-                                        jesUncertainty] = met_T1_py_jesDown[
-                                            jesUncertainty] - (
-                                                jet_pt_jesDownT1[jesUncertainty]
-                                                - jet_pt_L1) * jet_sinPhi
+                            # Calculate JES uncertainties on unsmeared MET
+                            for jesUncertainty in self.jesUncertainties:
+                                met_T1_px_jesUp[
+                                    jesUncertainty] = met_T1_px_jesUp[
+                                        jesUncertainty] - (
+                                            jet_pt_jesUpT1[jesUncertainty]
+                                            - jet_pt_L1) * jet_cosPhi
+                                met_T1_py_jesUp[
+                                    jesUncertainty] = met_T1_py_jesUp[
+                                        jesUncertainty] - (
+                                            jet_pt_jesUpT1[jesUncertainty]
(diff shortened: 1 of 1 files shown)


pieterdavid:
If I may take this opportunity to ask a question about https://github.com/cms-nanoAOD/nanoAOD-tools/pull/244: I noticed that all variations are different beween T1 and T1Smear, except for the unclustered energy, which is still a shift applied to the unsmeared (corrected) T1MET. Is this supposed to be also included when using T1Smear (and is it consistent then)? Thanks in advance!

pieterdavid:
ping. Could someone have a look? I may be wrong, but I think that this is a fairly straightforward bugfix

zdemirag:
Adding @alpakpinar about the unclustered uncertainty, there is an other PR and discussion on going as well.

gouskos:
Hi @alefisico  @camclean 
(a) The PR looks good to me. Do you have any tests/validation that we can try before merging?
(b) Can you please follow-up on this:

> If I may take this opportunity to ask a question about #244: I noticed that all variations are different beween T1 and T1Smear, except for the unclustered energy, which is still a shift applied to the unsmeared (corrected) T1MET. Is this supposed to be also included when using T1Smear (and is it consistent then)?

Thank you!

alefisico:
> Adding @alpakpinar about the unclustered uncertainty, there is an other PR and discussion on going as well.

sorry are you talking about #253? I think that is only for `FatJet` no?

alefisico:
> Hi @alefisico @camclean
> (a) The PR looks good to me. Do you have any tests/validation that we can try before merging?
> (b) Can you please follow-up on this:
> 
> > If I may take this opportunity to ask a question about #244: I noticed that all variations are different beween T1 and T1Smear, except for the unclustered energy, which is still a shift applied to the unsmeared (corrected) T1MET. Is this supposed to be also included when using T1Smear (and is it consistent then)?
> 
> Thank you!

For (a), from what I understand in the PR, I think the variables itself should not change correct? Maybe @pieterdavid  can plot some of the variables with the fix and without the fix?

For (b) including MET conveners @alkaloge @mcremone.

alkaloge:
Hi,

I am not sure what " I noticed that all variations are different beween T1 and T1Smear, except for the unclustered energy" describes/means actually. @pieterdavid can you clarify ? 

PS. In any case, given that MET does not have a reco code, we (MET) have not touched any part of the code for a long as I am aware of.

alefisico:
@alkaloge yes, still most of the changes are done by users. We (JME/JMAR) are working on some validation and a more "controlled"  way of implement changes. Hopefully we got something in place soon.

pieterdavid:
> For (a), from what I understand in the PR, I think the variables itself should not change correct? Maybe @pieterdavid can plot some of the variables with the fix and without the fix?

@alefisico Attached [log file](https://github.com/cms-nanoAOD/nanoAOD-tools/files/5582565/scan_mastervsPR256_splitjerornot.log) shows the jesTotal(Up,Down) variation with splitJER on and off for 10 events (DY MC). With the code in the master branch the values differ between splitJER=True and splitJER=False, with the PR they are the same (and the same as in the master branch with splitJER=False).

> I am not sure what " I noticed that all variations are different beween T1 and T1Smear, except for the unclustered energy" describes/means actually. @pieterdavid can you clarify ?

@alkaloge It's the same question as in https://github.com/cms-nanoAOD/nanoAOD-tools/issues/259 : the current code makes MET_T1_* and MET_T1Smear_* branches for all variations (since https://github.com/cms-nanoAOD/nanoAOD-tools/pull/244), but only MET_pt_unclustEn(Up,Down) and MET_phi_unclustEn(Up,Down), which are calculated with by adding the MetUnclustEnUpDelta(X,Y) values to the nominal MET_T1 - should these (a) also be used as a variation on the smeared Type-1 MET (which seems a bit strange) (b) not be used then or (c) instead be calculated based on the MET_T1Smear (which seems more consistent, but requires a rename to MET_T1_MetUnclustEn..., and the addition of MET_T1Smear_MetUnclustEn)?

alkaloge:
thanks, @pieterdavid now I see the problem. Indeed, I would go with c) (although a) vs c) should be close, but maybe not identical). Please keep in mind that in general MET does not really encourage people to use SmearMET as I don't think we have many use cases nor enough feedback on T1+Smear data/mc comparison.

pieterdavid:
Thanks @alkaloge ! They are close indeed, I implemented the change to check, and pushed it as a separate PR https://github.com/cms-nanoAOD/nanoAOD-tools/pull/266

pieterdavid:
I think I answered all the questions and comments two weeks ago, but please let me know if anything is missing.

PR #265, opened 2020-11-18
Changes to esh rob

GluonicPenguin:
Changes are designed to handle the eleven JES sources now produced in the HToInv-NanoaodTools repo, according to the reduced set of JES sources here: https://twiki.cern.ch/twiki/bin/viewauth/CMS/JECUncertaintySources


eshwen:
You need to make this PR to my fork: https://github.com/eshwen/nanoAOD-tools

PR #267, opened 2020-11-25
Two technical updates

hqucms:
This PR introduces two technical updates:

 - https://github.com/cms-nanoAOD/nanoAOD-tools/commit/0ac785c849d7d4cac327265a423e41414603874c changes the code to raise an error when reading a TTree branch on which `SetBranchStatus(name, 0)` has been called (i.e., when they are dropped in the `keep_and_drop` file). Otherwise the branch reading fails silently, leading to hard-to-detect errors.

- https://github.com/cms-nanoAOD/nanoAOD-tools/commit/bd96eb46171a9ed0824750f65c7fe949e69ac858 fixes a crash in python 3.7+ due to changes in python's `re` package.

### DIFF
--- a/python/postprocessing/framework/branchselection.py
+++ b/python/postprocessing/framework/branchselection.py
@@ -1,4 +1,9 @@
 import re
+try:
+    Pattern = re._pattern_type
+except AttributeError:
+    # Python 3.7
+    Pattern = re.Pattern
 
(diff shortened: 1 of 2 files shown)


hqucms:
ping @gouskos @mariadalfonso

PR #289, opened 2021-09-05
Add UL tau ID SFs

IzaakWN:
Add UL tau ID SFs, as presented in https://indico.cern.ch/event/1071293/#13-tau-sfs-for-ul


mariadalfonso:
as note: nanoTools should be updated to use the json files

IzaakWN:
Hi @mariadalfonso, yes, I agree. For now it seemed easier to quickly update the old ROOT files & python tool. Implementing `correctionlib` might require a bit more careful thought and work.

PR #292, opened 2021-10-11
Add Summer19UL17_JRV3_MC tarball and update jerTagsMC for UL2017

nurfikri89:
Adding `Summer19UL17_JRV3_MC.tar.gz` tarball and update UL2017  jerTagsMC in `jetmetHelperRun2.py`. This is needed because in `Summer19UL17_JRV2_MC.tar.gz`, there is _only_ the AK4PFchs text files and if one uses `jetmetHelperRun2.py` for AK8 jet which sets up `fatJetUncertainties.py`, it will crash for UL2017.

The `Summer19UL17_JRV3_MC.tar.gz` tarball was made from cms-jet/JRDatabase [repo](https://github.com/cms-jet/JRDatabase/tree/master/textFiles/Summer19UL17_JRV3_MC).


nurfikri89:
@gouskos I fully agree

bartokm:
This is strange. Did you make the tarball from the mentioned txt files? There is no JRV3 tarball for UL17 here: https://github.com/cms-jet/JRDatabase/tree/master/tarballs
I've asked about this ~1 month ago in the [JETMET HN](https://hypernews.cern.ch/HyperNews/CMS/get/JetMET/2146/1.html) and they replied that there's only correction for the AK4PFchs and they didn't announce deriving new corrections, neither the txt files are new which you linked.

kirschen:
V2 is still the official version, cf. https://twiki.cern.ch/twiki/bin/view/CMS/JetResolution#2017_data and V3 only used for internal studies so far. Better to stay with V2 for the time being in nano-tools to make sure this is consistent with released versions.

nurfikri89:
@kirschen noted. Will close this PR

PR #302, opened 2022-08-21
merge THnSparse object in haddnano

diogodebastos:
A THnSparse object is needed to save the models in pMSSM. For other analysis using nanoAOD tools within SUSY that are involved in the pMSSM paper this will be needed.


clelange (approved):
Thanks for this!

PR #304, opened 2023-03-03
Added check for negative indices in Collection class

meisonlikesicecream:
I added a check for negative indices when accessing an element in a Collection, so now it throws and error if a negative index is used.


clelange (approved):
Thanks for continuing to contribute improvements, looks good to me!

PR #306, opened 2023-06-15
Protect against files with zero pre-selected events

IzaakWN:
This PR adds a simple if-statement to protect to the case were zero events were preselected.

I ran into this error for several data files for which no events were in the good event list of the golden JSON, leading to failing jobs (`self.noOut` set to `True`). Otherwise I have to manually filter the input file list for the jobs.

### DIFF
--- a/python/postprocessing/framework/postprocessor.py
+++ b/python/postprocessing/framework/postprocessor.py
@@ -238,7 +238,7 @@ class PostProcessor:
                     eventRange=eventRange, maxEvents=self.maxEntries
                 )
                 print('Processed %d preselected entries from %s (%s entries). Finally selected %d entries' % (nall, fname, nEntries, npass))
-            elif outTree!=None:
+            elif outTree is not None:
                 nall = nEntries
                 print('Selected %d / %d entries from %s (%.2f%%)' % (outTree.tree().GetEntries(), nall, fname, outTree.tree().GetEntries() / (0.01 * nall) if nall else 0))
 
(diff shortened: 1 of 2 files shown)


clelange:
Thanks! So https://github.com/cms-nanoAOD/nanoAOD-tools/pull/305/ didn't solve this issue but this is about empty trees?

IzaakWN:
Thanks for the follow up.

No, I have the latest version and my jobs crash for files with zero pre-selected events because `outTree` is `None`.

I also use `noOut` set to true, otherwise, if `noOut` is false it would still fail few lines down when trying to write:
https://github.com/cms-nanoAOD/nanoAOD-tools/pull/306/files#diff-7a03b4209a618c9eb78be60742e95aa26c10bcdb784ef7f22b8b37550a79ff9eL247

Not sure if we prefer a totally empty output file?

python/postprocessing/framework/postprocessor.py
```diff
@@ -238,7 +238,7 @@ def run(self):
                     eventRange=eventRange, maxEvents=self.maxEntries
                 )
                 print('Processed %d preselected entries from %s (%s entries). Finally selected %d entries' % (nall, fname, nEntries, npass))
-            else:
+            elif outTree!=None:
```
meisonlikesicecream, line 241: `elif outTree is not None:`? :-)

meisonlikesicecream:
> @meisonlikesicecream let us know if you have any comments and/or you're OK with this, too, mainly to avoid duplications. Then I would merge.

@clelange I'm OK, I always have an output tree, so this shouldn't affect me

python/postprocessing/framework/postprocessor.py
IzaakWN: Done. The rest rest of the code indeed uses `is`.

IzaakWN:
I implemented @meisonlikesicecream's style suggestion, and sneaked in small cosmetic change for `haddnano.py` (adding spaces in print out).

meisonlikesicecream (approved):

meisonlikesicecream:
Aww I can't review it, but looks good to me (nice with the spaces 😄)

clelange (approved):

PR #301, opened 2022-05-26
Add statusflag help function to Object class

IzaakWN:
This small class method would help simplify some code and make it more readable:
```
for part in Collection(event,'GenPart'):
  isPrompt = part.statusflag('isPrompt')
  isPromptHard = part.statusflag('isPrompt','fromHardProcess')
  # ...
```
I left the `statusflags` dictionary global in `datamodel.py` so the user can still update the dictionary if needed in the future, or import it for some other purpose
```
import PhysicsTools.NanoAODTools.postprocessing.framework.datamodel as datamodel
datamodel.statusflags['fromHardProcess'] = 8
datamodel.statusflags['hard'] = 8 # shorthand
```

PR #307, opened 2023-08-08
Fix ord bug in prefetch for python2-3 compatibility

IzaakWN:
# BUG 1: `ord`
This bug appears when using the prefetch option in python3:
```
$ python2 -c "import os; print(type(os.urandom(8))); print(ord(os.urandom(8)[1]))"
<type 'str'>
106

$ python3 -c "import os; print(type(os.urandom(8))); print(ord(os.urandom(8)[1]))"
<class 'bytes'>
  File "<string>", line 1, in <module>
TypeError: ord() expected string of length 1, but int found
```
For python3, `ord` should be removed for this to work.

The best solution for keeping compatibility between both python2 and python3 I could immediately find is using `bytearray`
```
$ python2 -c 'import os; a=os.urandom(8); print(", ".join([hex(ord(i))[2:] for i in a])); print(", ".join([hex(i)[2:] for i in bytearray(a)]))'
38, 6e, b3, d2, fb, 70, 4d, e6
38, 6e, b3, d2, fb, 70, 4d, e6

$ python3 -c 'import os; a=os.urandom(8); print(", ".join([hex(i)[2:] for i in a])); print(", ".join([hex(i)[2:] for i in bytearray(a)]))'
5d, 42, 7, 61, 5a, 16, bd, 56
5d, 42, 7, 61, 5a, 16, bd, 56
```

# BUG 2: Encoding for hashing
This bug appears when using the `prefetch` _and_ the `longTermCache`  option in python3:
```
$ python2 -c 'import os, hashlib; print(os.getuid()); print(hashlib.sha1("fname").hexdigest())'
628
6dbff8baa2167c52eeb72d407642eb430ca0b271

$ python3 -c 'import os, hashlib; print(os.getuid()); print(hashlib.sha1("fname").hexdigest())'
628
  File "<string>", line 1, in <module>
TypeError: Unicode-objects must be encoded before hashing
```
Simplest fix that works for both python2 and python3 is to add `.encode("utf-8")`:
```
$ python2 -c 'import os, hashlib; print(os.getuid()); print(hashlib.sha1("fname".encode("utf-8")).hexdigest())'
628
6dbff8baa2167c52eeb72d407642eb430ca0b271

$ python3 -c 'import os, hashlib; print(os.getuid()); print(hashlib.sha1("fname".encode("utf-8")).hexdigest())'
628
6dbff8baa2167c52eeb72d407642eb430ca0b271
```

### DIFF
--- a/python/postprocessing/framework/postprocessor.py
+++ b/python/postprocessing/framework/postprocessor.py
@@ -65,7 +65,7 @@ class PostProcessor:
             return fname, False
         rndchars = "".join([hex(i)[2:] for i in bytearray(os.urandom(8))]) \
             if not self.longTermCache else "long_cache-id%d-%s" \
-            % (os.getuid(), hashlib.sha1(fname).hexdigest())
+            % (os.getuid(), hashlib.sha1(fname.encode('utf-8')).hexdigest())
         localfile = "%s/%s-%s.root" \
             % (tmpdir, os.path.basename(fname).replace(".root", ""), rndchars)
         if self.longTermCache and os.path.exists(localfile):
--- a/python/postprocessing/framework/postprocessor.py
+++ b/python/postprocessing/framework/postprocessor.py
@@ -63,8 +63,8 @@ class PostProcessor:
         tmpdir = os.environ['TMPDIR'] if 'TMPDIR' in os.environ else "/tmp"
         if not fname.startswith("root://"):
             return fname, False
-        rndchars = "".join([hex(ord(i))[2:] for i in os.urandom(
-            8)]) if not self.longTermCache else "long_cache-id%d-%s" \
+        rndchars = "".join([hex(i)[2:] for i in bytearray(os.urandom(8))]) \
+            if not self.longTermCache else "long_cache-id%d-%s" \
             % (os.getuid(), hashlib.sha1(fname).hexdigest())
         localfile = "%s/%s-%s.root" \
             % (tmpdir, os.path.basename(fname).replace(".root", ""), rndchars)


namapane:
Hi @IzaakWN 
Please not that the postprocessor is being imported in CMSSW at the moment:
https://github.com/cms-sw/cmssw/pull/42323
Both items were addressed therein; although without an attempt to preserve py2 compatibility (since this PR will be included in py3 based releases only).

namapane:
OK; since https://github.com/cms-sw/cmssw/pull/42323 is still pending, my idea would be to have #301 merged here, and ported in 42323, so that there's no need for a separate PR, review and approval.
Regarding the present PR, it can surely be merged as well once you validate it in py2. 
The implementation in 42323 does not incluide the bytearray trick to keep compatibility with py2, I'm not sure it is worth including it as I mentioned, although that would not hurt.

IzaakWN:
All good for me.

One other suprising bug that came up for me is that while testing the other PR https://github.com/cms-nanoAOD/nanoAOD-tools/pull/301 in python2 via CMSSW_10_3_3 and CMSSW_11_2_5, I got this error:
```
particles = Collection(event,'GenPart')
... len(particles) ...
...
TypeError: __len__() should return an int
```
because in this line in of `Collections.__init__`, the value for `nGenPart` (stored as `UInt_t` in nanoAOD) is of the type `long`...
https://github.com/cms-nanoAOD/nanoAOD-tools/blob/d2398cb1dfc5cd48e8b120e6469c8fc668cc263a/python/postprocessing/framework/datamodel.py#L116
This is strange to me that I have never run into this error before, but it's not due to any recent changes. It's just that no one seems to use `len(Collection)`... 😅 A simple solution is to use `Collection._len` instead.

It works in python3 as expected.

namapane:
> All good for me.
> 
> One other suprising bug that came up for me is that while testing the other PR #301 in python2 via CMSSW_10_3_3 and CMSSW_11_2_5, I got this error:
> 
> ```
> particles = Collection(event,'GenPart')
> ... len(particles) ...
> ...
> TypeError: __len__() should return an int
> ```
> 
> because in this line in of `Collections.__init__`, the value for `nGenPart` (stored as `UInt_t` in nanoAOD) is of the type `long`...
> 
> https://github.com/cms-nanoAOD/nanoAOD-tools/blob/d2398cb1dfc5cd48e8b120e6469c8fc668cc263a/python/postprocessing/framework/datamodel.py#L116
> 
> This is strange to me that I have never run into this error before, but it's not due to any recent changes. It's just that no one seems to use `len(Collection)`... 😅 A simple solution is to use `Collection._len` instead.
> 
> It works in python3 as expected.

That's strange to me as well since I have been definitely using len(collection) in CMSSW_10_6 and I never ran into this problem... ?

IzaakWN:
It does not seem related to the nanoAOD version (I tried older UL2018, and pre-UL files).

> That's strange to me as well since I have been definitely using len(collection) in CMSSW_10_6 and I never ran into this problem... ?

I think I never noticed it because only used `len()` on a (filtered) list of objects that was filled in a loop over a `Collection`. Perhaps the same for you?

clelange:
Thanks for this, I think it's still important to support both Python 2 and 3 since a lot of people will work in CMSSW releases that do not (yet) support Python 3. And for integration in CMSSW, we might need backports to those release lines.

@namapane Please let me know if you have any concerns. If not, i would merge it and then also https://github.com/cms-nanoAOD/nanoAOD-tools/pull/301 -- all these changes could then go into the CMSSW PR.

namapane:
> Thanks for this, I think it's still important to support both Python 2 and 3 since a lot of people will work in CMSSW releases that do not (yet) support Python 3. And for integration in CMSSW, we might need backports to those release lines.
> 
> @namapane Please let me know if you have any concerns. If not, i would merge it and then also #301 -- all these changes could then go into the CMSSW PR.

I agree that this repo must preserve py2 compatibility for future reference; I am OK with merging this and also #301.
Thanks,
N.

clelange (approved):

namapane:
> 



> It does not seem related to the nanoAOD version (I tried older UL2018, and pre-UL files).
> 
> > That's strange to me as well since I have been definitely using len(collection) in CMSSW_10_6 and I never ran into this problem... ?
> 
> I think I never noticed it because only used `len()` on a (filtered) (plain python) list of objects that was filled in a loop over a `Collection`. Perhaps the same for you?

Uhm, I can't reproduce this issue.
```
        genpart=Collection(event,"GenPart")
        print(len(genpart))
```
works for me both in py2 and py3...
Do you have a minimal script to reproduce this issue? (we should maybe move the discussion out of this PR since it's unrelated)

